PR #8, opened 2026-05-14
Update Gemini model name

### LINKED ISSUE #6: Change Gen Model Name
Please update the model name as follows:

Current model: gemini-1.5-pro (this version is deprecated)  
Replace with: gemini-2.5-flash

mzl2233:
This updates the Gemini model configuration from the deprecated `gemini-1.5-pro` value to `gemini-2.5-flash` in both the TypeScript source and the committed build output so the runtime uses the current model name consistently. Fixes #6.


ronisarkarexe:
@mzl2233 Thanks for the PR.

PR #383, opened 2026-05-24
fix: allow spaces in signup name field

### LINKED ISSUE #356: BUG: Signup 'Name' field restricts spaces and rejects full names

### What happened?

The signup page name input field does not allow users to enter spaces between words.

For example, users cannot enter full names like:

* Harendra Godara
* John Doe

The field currently only accepts letters, numbers, underscores, and dots.

### What did you expect, and what happened instead?

The name field should allow spaces so users can enter their real full names properly during signup.

Instead, spaces are rejected by the validation system, which creates a poor user experience.

### Steps to reproduce

1. Open the signup page
2. Click on the name input field
3. Try entering a full name with spaces (e.g. "Harendra Godara")
4. Observe that spaces are not accepted in the input field

### Environment (if relevant)

* OS: Fedora Linux
* Browser: Google Chrome
* Environment: Local development setup
* Frontend: React + Vite

N/A — A screenshot is not strictly required because the issue can be reproduced easily using the signup form validation.

### Record

SOHALIYAJAY:
## Description

Fixes #356  by updating the signup Name field validation to allow spaces in full names.

Previously, the frontend validation regex rejected whitespace characters, preventing users from entering names such as:

* John Doe
* Harendra Godara
* Jay Sohaliya

This PR updates the validation pattern to support spaces while preserving the existing restrictions against invalid/special characters.

---

### Frontend

* Updated the Name field validation regex in:

  * `frontend/src/components/signup/signup.component.tsx`
* Updated validation error message to reflect the new allowed input format.

#### Before

```regex
^[A-Za-z0-9._]+$
```

#### After

```regex
^[A-Za-z0-9\s._]+$
```

This change allows whitespace (`\s`) while still restricting unsupported special characters.

---

## Unchanged Behavior

The following validation rules remain unchanged:

* Required field validation
* Minimum length validation (`minLength: 8`)
* Email/password validation
* OTP flow and password strength checks

Backend validation was not modified because the issue was limited to frontend pattern validation.

---

### Valid Inputs

* Harendra Godara
* John Doe
* Jay Sohaliya
* A B

### Invalid Inputs

* @@@
* ###$$$
* name!!!
* `<script>`

Additionally:

* TypeScript checks completed successfully
* No unrelated files or logic were modified

## Before Fix

image

## After Fix

image

## Signup Flow After Fix
image

## Result

Users can now enter proper full names with spaces during signup without triggering validation errors, while invalid characters continue to be blocked.

### DIFF
--- a/frontend/src/components/signup/signup.component.tsx
+++ b/frontend/src/components/signup/signup.component.tsx
@@ -186,9 +186,9 @@ const SignUpComponent = () => {
                     message: "Name must be at least 8 characters",
                   },
                   pattern: {
-                    value: /^[A-Za-z0-9._]+$/,
+                    value: /^[A-Za-z0-9\s._]+$/,
                     message:
-                      "Only letters, numbers, underscores, and dots are allowed",
+                      "Only letters, numbers, spaces, underscores, and dots are allowed",
                   },
                 }}
                 error={errors.name}


SOHALIYAJAY:
Implemented the requested fix for issue #356 and verified the validation behavior locally. Please let me know if any further changes are needed.

PR #419, opened 2026-05-24
fix(backend): prevent concurrent AI generation quota bypass 

### LINKED ISSUE #357: BUG: AI generation request limits can be bypassed using concurrent requests

### What happened?

While reviewing the AI story generation flow, I noticed that the current request limit system can be bypassed when multiple requests hit the generation endpoint at the same time.

Right now the request limit is checked before the slow Gemini generation process starts, but the database counter is only updated after the generation successfully completes.

Because of this, several concurrent requests can pass the same limit check before the counter is incremented.

I also noticed that the current timeout implementation uses `Promise.race`, but the pending Gemini request is not actually cancelled after timeout. This means timed-out requests may continue running in the background and still consume external API resources.

This can lead to:

* request quota bypass
* inconsistent `requestsThisMonth` values
* unnecessary Gemini API usage/billing
* background resource leakage during timeouts

### Steps to reproduce

1. Create a user account with a low remaining monthly generation limit.
2. Send multiple concurrent requests to `/generate-model` using `Promise.all` or parallel API calls.
3. Observe that several requests pass the limit check simultaneously before the database counter updates.
4. Verify that more stories are generated than the allowed quota.
5. Optionally simulate slow generation requests or timeouts and observe that external API operations continue running in the background.

### Environment (if relevant)

* Backend: Node.js + Express + MongoDB
* AI Service: Gemini API
* Environment tested locally on Linux
* API Surface: `/generate-model` and `/generate-free-model`

_No response_

### Record

KRUSHAL2956:
## 📌 Description

This PR fixes a critical concurrency vulnerability in the AI story generation flow where users could bypass monthly generation limits by sending multiple requests simultaneously.

Previously, the request limit validation happened before the slow Gemini generation process started, while the database counter was only updated after generation completed successfully. Because of this read-before-write flow, multiple concurrent requests could pass the same limit check before the counter updated.

This implementation refactors the quota flow into a concurrency-safe atomic reservation system using MongoDB atomic updates.

---

## 🔗 Related Issue

Closes #357

---

### Atomic Quota Reservation
- Refactored authenticated AI generation flow to use atomic `findOneAndUpdate` quota reservation
- Removed unsafe read-before-write request validation logic
- Ensured quota reservation happens before Gemini generation begins

### Rollback & Failure Safety
- Added rollback protection to refund quota if Gemini generation fails or times out
- Added defensive safeguards to prevent negative counters during rollback

### Middleware Cleanup
- Simplified `check.request.limit.ts` into a strict authentication-only middleware
- Removed duplicate and unsafe quota validation logic from middleware layer

### Free Generation Hardening
- Refactored anonymous/free generation flow so counters increment synchronously before async execution
- Prevented concurrent bypass in free generation mode using event-loop-safe state locking

### Stability & Concurrency Improvements
- Eliminated race conditions during simultaneous generation bursts
- Prevented inconsistent `requestsThisMonth` states under parallel requests
- Reduced risk of unnecessary Gemini API usage during concurrent abuse attempts

---

### Manual Concurrency Validation

Tested concurrent generation bursts using parallel requests:

```ts
Promise.all([...Array(20)].map(() => generateStory()))

### DIFF
--- a/backend/package.json
+++ b/backend/package.json
@@ -44,6 +44,7 @@
     "@types/node-cron": "^3.0.11",
     "@types/nodemailer": "^6.4.17",
     "@types/socket.io": "^3.0.1",
+    "@types/winston": "^2.4.4",
     "ts-node": "^10.9.2",
     "ts-node-dev": "^2.0.0",
     "typescript": "~5.7.2"
--- a/backend/src/app/middleware/check.request.limit.ts
+++ b/backend/src/app/middleware/check.request.limit.ts
@@ -28,29 +28,6 @@ const checkRequestLimit =
         throw new ApiError(httpStatus.BAD_REQUEST, "User not found!");
       }
 
-      const currentDate = new Date();
-      const firstDayOfMonth = new Date(
-        currentDate.getFullYear(),
-        currentDate.getMonth(),
-        1
-      );
-
-      // Reset the request count if the last request was in a previous month
-      if (user.lastRequestDate && user.lastRequestDate < firstDayOfMonth) {
-        user.requestsThisMonth = 0;
-        user.lastRequestDate = currentDate;
-      }
-
-      const requestLimit =
-        REQUEST_LIMITS[user.subscriptionType as keyof typeof REQUEST_LIMITS];
-
-      // Check if the user has exceeded their monthly limit
-      if (user.requestsThisMonth >= requestLimit) {
-        throw new ApiError(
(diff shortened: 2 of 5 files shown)


ronisarkarexe:
@KRUSHAL2956 Please give a star to support my project.

KRUSHAL2956:
yes sure.

PR #459, opened 2026-05-25
fix: prevent password re-hashing in user pre-save hook

### LINKED ISSUE #454: BUG: [Security] Fix password re-hashing bug in User model pre-save hook

## Describe the bug

The current password hashing logic inside the User model `pre("save")` hook re-hashes the password every time the user document is saved, even when the password itself was not modified.

Current implementation:

```ts
UserSchema.pre("save", async function (next) {
  const user = this;

  if (user.password && user.password.trim() !== "") {
    user.password = await bcrypt.hash(
      user.password,
      Number(config.bcrypt_salt_rounds)
    );
  }

  next();
});

The condition only checks whether user.password exists, which is always true after initial signup because the hashed password is already stored in the document.

As a result:

updating unrelated fields (bio, avatar, followers, postsCount, etc.)
triggers another bcrypt hash operation,
the already-hashed password gets hashed again,
and future login attempts may fail because the stored password becomes a "hash of a hash."

This can lead to:

user account lockout,
authentication inconsistencies,
unnecessary CPU-intensive hashing operations,
and backend performance waste.


Expected behavior

Passwords should only be hashed when the password field itself is modified.

Suggested fix:

if (!user.isModified("password")) {
  return next();
}

This ensures:

new passwords are hashed correctly,
existing hashes are preserved during unrelated profile updates,
and authentication integrity remains stable.
Actual behavior

The password hashing middleware executes during every save() operation regardless of whether the password was changed.




*File Path: backend/src/app/modules/user/user.model.ts
* Labels->type:bug, type:security, type:performance

### Steps to reproduce



# Steps to Reproduce
1. Register a new user account with a password
   Example:
   password = "hello123"

2. Verify that login works correctly

3. Update any unrelated user field
   Examples:
   - bio
   - avatar
   - followers
   - postsCount

4. Trigger `user.save()`

5. Observe that the pre-save hook runs again and hashes the already-hashed password

6. Attempt login again using the original password

7. Observe authentication failure or inconsistent password comparison behavior

### Environment (if relevant)

_No response_

_No response_

### Record

Priya09023:
## Description

This PR fixes a security and performance issue in the User model pre-save hook where passwords were being re-hashed on every `save()` operation, even when the password field was not modified.

### Changes made

* Added `user.isModified("password")` check before hashing
* Prevented already-hashed passwords from being hashed again
* Preserved existing password hashes during unrelated profile updates

### Impact

* Fixes potential login failures caused by hash-of-hash passwords
* Reduces unnecessary bcrypt operations
* Improves authentication stability

Fixes #454

### DIFF
--- a/backend/src/app/modules/user/user.model.ts
+++ b/backend/src/app/modules/user/user.model.ts
@@ -62,8 +62,8 @@ export const UserSchema: Schema<IUser> = new Schema<IUser, UserModel>(
 UserSchema.pre("save", async function (next) {
   const user = this;
   if (!user.isModified("password")) {
-  return next();
-}
+    return next();
+  }
   
   // Only hash password if it exists and is not empty (for password-based auth)
   // Skip for Google OAuth users who don't have passwords
--- a/backend/src/app/modules/user/user.model.ts
+++ b/backend/src/app/modules/user/user.model.ts
@@ -61,6 +61,9 @@ export const UserSchema: Schema<IUser> = new Schema<IUser, UserModel>(
 
 UserSchema.pre("save", async function (next) {
   const user = this;
+  if (!user.isModified("password")) {
+  return next();
+}
   
   // Only hash password if it exists and is not empty (for password-based auth)
   // Skip for Google OAuth users who don't have passwords


ronisarkarexe:
@Priya09023 Please give a star to support my project.

PR #504, opened 2026-05-25
perf: optimize dashboard analytics queries using database aggregation

### LINKED ISSUE #503: [Performance] Dashboard analytics performs large in-memory filtering instead of database-side aggregation

While exploring the dashboard analytics service, I noticed that several analytics calculations currently fetch complete collections from MongoDB and process them in server memory using JavaScript filtering.

Example from analysis.service.ts:

const users = await User.find({});
const activeUsers = users.filter((u) => u.status === USER_STATUS.ACTIVE).length;

A similar pattern is also used for post statistics, subscription counts, and monthly grouping logic.

Suggested Improvement

Move these calculations to the database layer using:

countDocuments()
aggregation pipelines
grouped queries

For example, monthly post statistics could be calculated directly in MongoDB using $group instead of processing all records in memory.

Expected Benefits

This would:

reduce server memory usage
improve dashboard response performance
make analytics queries more scalable
reduce unnecessary data transfer
improve overall backend efficiency
Possible Scope
user statistics
subscription analytics
post counts
monthly grouping metrics
dashboard summary cards
Impact

This would improve the backend architecture significantly and help the analytics module scale much better as the platform grows.

### Why does this matter?

Why This Is a Problem

This approach works for small datasets, but as the number of users and stories grows, it can become a serious scalability bottleneck.

Potential issues:

unnecessary memory usage on the Node.js server
blocking of the event loop during large array operations
slower dashboard response times
increased MongoDB → Node transfer overhead
higher risk of performance degradation under load

_No response_

### Record

premdeshmane-01:
## Description

This PR improves the performance and scalability of the dashboard analytics module by replacing large in-memory filtering operations with database-side aggregation and counting queries.

### Changes Made

* Replaced collection-wide `find({})` fetching patterns
* Optimized analytics calculations using:

  * `countDocuments()`
  * MongoDB aggregation pipelines
* Reduced unnecessary in-memory array filtering
* Improved query efficiency for dashboard statistics

### Why This Fix Is Needed

Previously, the analytics service fetched complete MongoDB collections into Node.js memory and processed them using JavaScript filtering logic.

This approach can become inefficient as the dataset grows because it:

* increases server memory usage
* blocks the Node.js event loop during heavy filtering
* increases unnecessary database-to-server data transfer
* slows dashboard response times under load

### Improvements

This update:

* shifts computation to MongoDB
* improves backend scalability
* reduces server-side memory overhead
* improves dashboard responsiveness
* prepares the analytics module for larger production datasets

## Related Issue

Fixes #503

## Type of Change

* [x] Performance improvement
* [x] Backend optimization
* [ ] New feature
* [ ] Documentation update
* [ ] Code refactor

## Tested

* Verified analytics endpoints return expected values
* Confirmed aggregation queries work correctly
* Tested dashboard analytics locally after optimization

## Checklist

* [x] I tested my changes locally
* [x] I performed a self-review
* [x] Only relevant backend files were modified

### DIFF
--- a/backend/src/app/modules/analysis/analysis.service.ts
+++ b/backend/src/app/modules/analysis/analysis.service.ts
@@ -1,61 +1,90 @@
 import { SUBSCRIPTION_TYPE } from "../../../enums/subscription_type";
 import { ENUM_USER_ROLE } from "../../../enums/user";
 import { USER_STATUS } from "../../../enums/user_status";
-import { IPost } from "../post/post.interface";
 import { Post } from "../post/post.model";
 import { User } from "../user/user.model";
 
 const getDashboardAnalysis = async () => {
-  // get all users
-  const users = await User.find({});
-  const totalUsers = users.length;
-  const activeUsers = users.filter(
-    (u) => u.status === USER_STATUS.ACTIVE
-  ).length;
-  const inactiveUsers = users.filter(
-    (u) => u.status === USER_STATUS.INACTIVE
-  ).length;
-  const blockedUsers = users.filter(
-    (u) => u.status === USER_STATUS.BLOCKED
-  ).length;
-  const writers = users.filter((u) => u.role === ENUM_USER_ROLE.WRITER).length;
-  const applyForWriter = users.filter(
-    (u) => u.isApplyForWriter === true
-  ).length;
+  // Run all database queries in parallel for maximum throughput.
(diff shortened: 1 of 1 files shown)


ronisarkarexe:
@premdeshmane-01 Please give a star to support my project.

premdeshmane-01:
@ronisarkarexe Already starred!!
keep growing Brother !

PR #650, opened 2026-05-27
fix: cache model and tokenizer at module level to avoid repeated disk I/O

### LINKED ISSUE #639: bug: scorer.py loads model and tokenizer on every API call causing performance bottleneck
## Problem

In `backend/ml/scorer.py`, the `score()` function loads the model 
and tokenizer from disk on every single call:

    model     = load_model(model_path)
    tokenizer = joblib.load(tokenizer_path)

This means every time a story is scored via `POST /score`, the entire 
BiLSTM model is read from disk. In production, with multiple stories 
being scored per request, this causes severe latency.

## Expected Behavior

The model and tokenizer should be loaded once at module level and 
reused across all calls.

## Suggested Fix

Move model/tokenizer loading outside the function:

    # Module level — load once
    _model = load_model(os.path.join(SAVE_DIR, "scorer.keras"))
    _tokenizer = joblib.load(os.path.join(SAVE_DIR, "tokenizer.pkl"))

    def score(story_text: str, prompt_text: str) -> dict:
        # use _model and _tokenizer directly

## Files Affected
- `backend/ml/scorer.py`

## Labels
bug, backend, good first issue

shreyjeetdas:
## Screenshot (when helpful)
N/A — backend code change, no UI involved.

## What this PR does
In scorer.py, the score() function was loading the BiLSTM model and 
tokenizer from disk on every single API call. This means every POST 
/score request triggered a full model load, causing unnecessary latency 
especially when scoring multiple stories per request.

This PR moves model and tokenizer loading into a module-level 
_load_artifacts() function that uses _model and _tokenizer cache 
variables. The artifacts are loaded from disk only on the first call 
and reused for all subsequent calls.

## Type of change
- [x] Bug fix
- [ ] New feature
- [ ] Code refactor
- [ ] Documentation update

## Related issue
Fixes #639

N/A

## Checklist
- [ ] I have added screenshots or a recording when they help explain this PR, or noted N/A with a one-line reason.

### DIFF
--- a/backend/ml/scorer.py
+++ b/backend/ml/scorer.py
@@ -28,6 +28,9 @@ EMBED_DIM  = 64
 LSTM_UNITS = 64
 SAVE_DIR   = os.path.join(os.path.dirname(__file__), "saved_scorer")
 
+# Module-level cache — model and tokenizer loaded once on first call
+_model     = None
+_tokenizer = None
 
 def build_model() -> Model:
     """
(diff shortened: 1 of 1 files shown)


ronisarkarexe:
@shreyjeetdas Please give a star to support my project.

PR #817, opened 2026-05-28
fix: return not found for missing user delete

### LINKED ISSUE #812: Bug: deleteUser succeeds even when no user is deleted
Line: backend/src/app/modules/user/user.service.ts:66

I noticed this while reading through the user deletion service. `deleteUser` calls `User.deleteOne({ _id: id })` and then returns without checking whether MongoDB actually deleted anything.

Problem:
If the id is valid-looking but does not match any user, the service still completes successfully. The controller can report a successful delete even though no user was removed.

Impact:
Admin/user-management flows can show a false success state. This also makes debugging harder because a bad id and a real delete look the same to the caller.

A quick way to verify:
- Call the delete-user path with a valid ObjectId format that does not exist in the database.
- The service does not throw.
- The caller receives success even though `deletedCount` would be 0.

Expected behavior:
The service should throw a 404-style error when no document is deleted.

Suggested fix:
Store the `deleteOne` result and check `result.deletedCount`. If it is 0, throw `ApiError(httpStatus.NOT_FOUND, "User not found!")` or the project equivalent.

Labels: bug, gssoc'26, level:beginner

I searched existing open issues and did not find this reported. Raising this as part of GSSoC 2026.

- [x] I agree to follow this project Code of Conduct

eshaanag:
## Screenshot (when helpful)

N/A - this is a backend API error-handling fix and does not change the UI.

## What this PR does

This fixes the user deletion path reported in #812.

Before this change, `deleteUser` called `User.deleteOne({ _id: id })` and returned successfully without checking whether MongoDB deleted a document. A delete request for a valid-looking but nonexistent user id could therefore be reported as successful.

With this change:

- `deleteUser` checks the `deleteOne` result
- `deletedCount === 0` now throws `ApiError(httpStatus.NOT_FOUND, "User not found!")`
- the delete controller uses `catchAsync`, so the not-found error reaches the global error handler instead of being converted into a generic 400 response

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Code refactor
- [ ] Documentation update

## Related issue

Closes #812

N/A

## Checklist

- [ ] I have tested my changes.

## Local checks

- `git diff --check` passed.
- `npm run build --workspace backend` could not run because `tsc` is not installed in this checkout.
- `npm install` was already attempted in this workspace and failed with `ENOSPC`, so I could not install the missing backend toolchain locally.

### DIFF
--- a/backend/src/app/modules/user/user.controller.ts
+++ b/backend/src/app/modules/user/user.controller.ts
@@ -61,23 +61,17 @@ const updateUser = async (req: Request, res: Response) => {
   }
 };
 
-const deleteUser = async (req: Request, res: Response) => {
-  try {
-    const id = routeParam(req.params.id);
+const deleteUser = catchAsync(async (req: Request, res: Response) => {
+  const id = routeParam(req.params.id);
 
-    await UserService.deleteUser(id);
+  await UserService.deleteUser(id);
 
-    sendResponse(res, {
-      statusCode: httpStatus.OK,
-      success: true,
-      message: "User deleted successfully!",
-    });
-  } catch (error) {
-    res.status(httpStatus.BAD_REQUEST).json({
-      message: "Fail to get users!",
-    });
-  }
-};
+  sendResponse(res, {
+    statusCode: httpStatus.OK,
+    success: true,
+    message: "User deleted successfully!",
+  });
+});
 
 const applyForWriter = catchAsync(async (req: Request, res: Response) => {
   const token = await getToken(req);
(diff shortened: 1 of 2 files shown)


eshaanag:
Hi @maintainer, this PR was raised as part of GSSoC 2026 contribution. Could you please add the appropriate GSSoC label if applicable? Thanks!

eshaanag:
Resolved the merge conflict in `backend/src/app/modules/user/user.service.ts` and pushed the updated branch. I kept the upstream cleanup logic for deleting related user data, and added the missing not-found checks around user deletion.

The PR now shows as mergeable. The remaining CI failures appear to be from existing upstream files outside this PR diff, mainly `analytics.router.ts`, `recommendation.router.ts`, and existing frontend lint errors.

ronisarkarexe:
@eshaanag Please give a star to support my project.

PR #845, opened 2026-05-28
fix: remove redundant client-side tag filtering that conflicted with …

### LINKED ISSUE #844: Bug: Client-side tag filtering conflicts with server-side API filtering causing posts to disappear

## Bug Description
Posts are filtered twice — once server-side via the `genres` API param 
and again client-side using different string comparison logic. This 
causes valid posts returned by the API to be incorrectly hidden.

## Expected Behavior
Selecting a genre or tag should show all matching posts returned 
by the API.

## Actual Behavior
Some valid posts disappear after genre/tag selection because the 
client-side filter uses different logic than the API.

## Root Cause
`filteredPosts` applies a second `.filter()` on already-filtered 
API results:

```ts
// ❌ API already filters by genres, this causes conflict
const filteredPosts =
  selectedTags.length === 0
    ? data?.posts || []
    : (data?.posts || []).filter((post: Post) =>
        selectedTags.some(
          (selectedTag) =>
            post.tag?.toLowerCase().trim() ===
            selectedTag.toLowerCase().replace("#", "").trim(),
        ),
      );
```

```ts
// ✅ Fix — trust the API response
const filteredPosts = data?.posts || [];
```

## File Affected
`frontend/src/components/post/post.component.tsx`

## Steps to Reproduce
1. Go to `/explore`
2. Select any genre checkbox (e.g. Fantasy)
3. Some matching posts are missing from results

### Environment (if relevant)

_No response_

_No response_

### Record

akhilmodi29:
Fixes #844 

## Problem
Posts were being filtered twice — once by the API via the `genres` 
query param, and again on the client using different logic. This caused 
valid posts returned by the API to be incorrectly hidden from the user.

## Root Cause
The `filteredPosts` variable applied a second `.filter()` on top of 
already-filtered API results. The client filter compared `post.tag` 
against `selectedTag` using different string logic than the API, 
causing a mismatch.

## Fix
Removed the redundant client-side filter. `filteredPosts` now simply 
uses the API response directly since the server already handles 
genre/tag filtering correctly.

## Testing
- [ ] All posts load correctly with no filters applied
- [ ] Genre filter returns correct posts
- [ ] Tag filter returns correct posts
- [ ] Reset button restores all posts

## Type
- [x] Bug fix

### DIFF
--- a/frontend/src/components/post/post.component.tsx
+++ b/frontend/src/components/post/post.component.tsx
@@ -38,16 +38,7 @@ const ExploreComponent = () => {
 
   const { data, isLoading } = useGetPostListsQuery({ ...query });
 
-  const filteredPosts =
-    selectedTags.length === 0
-      ? data?.posts || []
-      : (data?.posts || []).filter((post: Post) =>
-          selectedTags.some(
-            (selectedTag) =>
-              post.tag?.toLowerCase().trim() ===
-              selectedTag.toLowerCase().replace("#", "").trim(),
-          ),
-        );
+  const filteredPosts = data?.posts || [];
 
   const resetAllStates = () => {
     setSortBy("createdAt");


ronisarkarexe:
@akhilmodi29 Please give a star to support my project.

PR #915, opened 2026-05-28
fix(config): throw at startup when DATABASE_URL is missing (closes #768)

### LINKED ISSUE #768: BUG: Hardcoded MongoDB Atlas credentials committed as fallback in config/index.ts

### What happened?

**File:** `backend/src/config/index.ts` — line 21

A real MongoDB Atlas connection string containing plaintext credentials is committed directly in the source code as a hardcoded fallback value:

```ts
database_url: process.env.DATABASE_URL || "mongodb+srv://university_system:Y4K8t26J0a894hke@cluster0.v4fkr.mongodb.net/ai_storie_books?retryWrites=true&w=majority",
```

Anyone who clones the repository has immediate visibility of the database username and password. Even if the Atlas cluster is now locked down or the password has been rotated, the pattern is dangerous — it teaches contributors that hardcoded credential fallbacks are acceptable, and any future rotation could be missed.

> **Note:** A related issue (#257, closed) addressed `console.log(config.database_url)` leaking the URI in startup logs. That fix removed the log statement but did **not** remove the hardcoded fallback string itself. This issue targets the root cause.

**Impact:** Full database read/write access for anyone who clones the repo. Credentials are permanently visible in git history.

### Steps to reproduce

1. Clone the repository
2. Open `backend/src/config/index.ts`
3. Line 21 contains a live Atlas URI with embedded username and password as the `||` fallback — no `.env` file needed to expose it

### Proposed fix

Remove the fallback entirely and throw a startup error if `DATABASE_URL` is not set:

```ts
database_url: (() => {
  if (!process.env.DATABASE_URL) {
    throw new Error("DATABASE_URL environment variable is required. See .env.example.");
  }
  return process.env.DATABASE_URL;
})(),
```

This ensures the server refuses to start without a properly configured environment, rather than silently falling back to committed credentials. The `.env.example` file already documents `DATABASE_URL` as a required variable, so this change is consistent with the project's existing setup guidance.

### Environment (if relevant)

- File: `backend/src/config/index.ts`
- Affects: all environments (dev, staging, production)
- No special setup needed to reproduce — the credentials are visible in the source file itself

N/A — the issue is visible directly in the source file.

### Record

nyxsky404:
## Before you open this PR

- **Issue:** Closes #768
- **Description:** Replaces the silent `undefined` fallback for `DATABASE_URL` with a startup-time throw, so the server refuses to boot without a properly configured environment.
- **Screenshots:** N/A — this is a backend config hardening change with no UI surface.

## Screenshot (when helpful)

N/A — backend-only change. The behavior is observable as a startup error in the terminal when `DATABASE_URL` is unset:

```
Error: DATABASE_URL environment variable is required. See backend/.env.example for setup instructions.
```

## What this PR does

Issue #768 flagged that `backend/src/config/index.ts` historically shipped a hardcoded MongoDB Atlas URI with embedded credentials as a fallback for `DATABASE_URL`. The hardcoded URI itself was removed in an earlier change, but the fallback was replaced with `|| undefined`, which lets the process boot in a half-configured state and defers the failure to `connectDB()` at request time.

This PR implements the fix proposed directly in the issue: an IIFE that throws synchronously at module load if `DATABASE_URL` is missing or empty. With this in place:

- The server cannot start without a valid `DATABASE_URL`. There is no path back to a hardcoded fallback, by construction.
- `config.database_url` is now typed as `string` instead of `string | undefined`, so consumers (`server.ts`, `scripts/seed-admin.ts`) no longer need a cast or a runtime nullish check.
- The previously redundant guard in `server.ts`'s `connectDB()` is removed, with a brief comment explaining why the guard is no longer needed.

The error message points contributors at `backend/.env.example`, which already documents `DATABASE_URL` as a required variable, so the change is consistent with the project's existing setup guidance.

### Follow-up actions for the maintainer (cannot be done in a PR)

The code change closes the root cause but does not, on its own, fully neutralize the original credential exposure. Two out-of-band steps remain:

1. **Rotate the leaked Atlas credentials.** The original username/password are still present in the git history of cloned repos. The exposed user should be rotated or deleted on the Atlas side.
2. **(Optional) Rewrite git history** with `git filter-repo` or BFG to purge the URI from past commits, then force-push and notify forks.

These require maintainer-level access to the Atlas console and the upstream repo, so they are flagged here rather than included in the PR.

## Type of change

Check all that apply (if more than one, explain in "What this PR does").

- [x] Bug fix
- [ ] New feature
- [ ] Code refactor
- [ ] Documentation update

## Related issue

Closes #768

N/A

## Checklist

### DIFF
--- a/backend/src/config/index.ts
+++ b/backend/src/config/index.ts
@@ -17,7 +17,15 @@ export default {
   env: process.env.NODE_ENV,
   port: process.env.PORT || "5000",
   disable_logs: process.env.DISABLE_LOGS === "true" || process.env.VERCEL === "1",
-  database_url: process.env.DATABASE_URL?.trim() || undefined,
+  database_url: (() => {
+    const url = process.env.DATABASE_URL?.trim();
+    if (!url) {
+      throw new Error(
+        "DATABASE_URL environment variable is required. See backend/.env.example for setup instructions."
+      );
+    }
+    return url;
+  })(),
   cors_origins: parseCorsOrigins(process.env.CORS_ORIGINS),
   bcrypt_salt_rounds: process.env.SALT_ROUNDS,
   jwt: {
--- a/backend/src/server.ts
+++ b/backend/src/server.ts
@@ -22,13 +22,9 @@ if (config.disable_logs) {
 
 async function connectDB() {
   if (mongoose.connection.readyState === 1) return;
-  const databaseUrl = config.database_url;
-  if (!databaseUrl) {
-    throw new Error(
-      "DATABASE_URL is not set. Define it in backend/.env before starting the server."
-    );
-  }
-  await mongoose.connect(databaseUrl);
+  // config.database_url is guaranteed non-empty by config/index.ts — it throws at
+  // module load time if DATABASE_URL is missing, so no runtime guard is needed here.
+  await mongoose.connect(config.database_url);
 }
 
 async function main() {


ronisarkarexe:
pls fix merge isssue

nyxsky404:
@ronisarkarexe , Fixed the merge conflicts

ronisarkarexe:
@nyxsky404 Please give a star to support my project.

PR #985, opened 2026-05-28
fix(backend): use replyTo instead of spoofed from header in contact email

### LINKED ISSUE #983: fix: use replyTo instead of spoofed from header in contact email to prevent SMTP authentication failures
### Bug Description
In `backend/src/utils/email.util.ts` (line ~70), the contact email transporter attempts to send the user's message using the user's email directly in the `from` header:

```ts
  const mailOptions = {
    from: `"${data.fullname}" <${data.email}>`,
    to: config.verify_email,
    subject: `Contact Form: ${data.subject}`,
    ...
```

### Why this is a Bug:
Most SMTP services (especially Gmail SMTP which is configured here via `service: "gmail"`) strictly forbid spoofing the `from` header. If the `from` field contains an email address (`data.email`) that does not match the authenticated user (`config.verify_email`), the SMTP server will either:
1. Reject the request entirely with an authentication error (e.g. `5.7.1 Sender address rejected: not owned by user` or `554 5.7.1 Message rejected under suspicion of SPAM`).
2. 2. Force-rewrite the `from` header back to the authenticated account's email, losing the sender's original display name and email.
### Proposed Solution:
To prevent SPF/DKIM validation failures and guarantee successful delivery, the email should be sent **from** the authenticated email address, with the `replyTo` header set to the sender's email. This allows the recipient to hit reply and email the user directly:

```ts
  const mailOptions = {
    from: `"${data.fullname}" <${config.verify_email}>`,
    replyTo: data.email,
    to: config.verify_email,
    subject: `Contact Form: ${data.subject}`,
    ...
```

As a contributor of **GirlScript Summer of Code (GSSoC) 2026**, I would love to solve this bug. Please assign it to me!

omermohammedfarooq:
## Description

Fixes #983

Most SMTP services (especially Gmail SMTP which is configured here via `service: "gmail"`) strictly forbid spoofing the `from` header. If the `from` field contains an email address (`data.email`) that does not match the authenticated user (`config.verify_email`), the SMTP server will either:
1. Reject the request entirely with an authentication error.
2. Force-rewrite the `from` header back to the authenticated account's email, losing the sender's original display name and email.

This PR sets the `from` header to the authenticated `config.verify_email`, and properly forwards the sender's original email in the `replyTo` header so that direct replies still reach the sender.

## Checklist

- [x] I have read the `CONTRIBUTING.md` rules.
- [x] I have tested this change locally.
- [x] My commits follow the Conventional Commits format.
- [x] The code is clean and contains no extra dependencies.

### DIFF
--- a/backend/src/utils/email.util.ts
+++ b/backend/src/utils/email.util.ts
@@ -67,7 +67,8 @@ export const sendContactEmail = async (data: {
   });
 
   const mailOptions = {
-    from: `"${data.fullname}" <${data.email}>`,
+    from: `"${data.fullname}" <${config.verify_email}>`,
+    replyTo: data.email,
     to: config.verify_email,
     subject: `Contact Form: ${data.subject}`,
     html: `


ronisarkarexe:
@omermohammedfarooq Please give a star to support my project.

PR #1065, opened 2026-05-29
fix: add startup guard for RAZORPAY_KEY_SECRET and use timingSafeEqual in verifyPayment

### LINKED ISSUE #1042: [Bug][Level 2] verifyPayment uses RAZORPAY_KEY_SECRET without a startup guard — missing env var causes opaque 500 errors instead of a clear configuration failure
## Summary

`verifyPayment` in `backend/src/controllers/payment.controller.ts` reads `process.env.RAZORPAY_KEY_SECRET` inline inside the handler and casts it with `as string` to silence TypeScript. When the environment variable is absent (a misconfigured staging deploy, a missing secret in CI, a fresh developer setup), `crypto.createHmac("sha256", undefined as any)` throws a `TypeError` at runtime. The catch block returns a generic 500 with the message `"Payment verification failed"`, giving operators no indication that the problem is a missing configuration value rather than a transient error.

## Affected File

`backend/src/controllers/payment.controller.ts`, lines 38-41

## Root Cause

```ts
const expectedSignature = crypto
  .createHmac("sha256", process.env.RAZORPAY_KEY_SECRET as string)
  .update(body)
  .digest("hex");
```

`process.env.RAZORPAY_KEY_SECRET` is `undefined` when the variable is not set. The `as string` cast tells TypeScript it is fine but does not change the runtime value. Node.js throws:

```
TypeError: The "key" argument must be of type string or an instance of ArrayBuffer, Buffer, TypedArray, DataView, KeyObject, or CryptoKey.
```

This error is swallowed by the `catch` block and returned as a 500.

## Impact

- Every payment verification call fails silently when `RAZORPAY_KEY_SECRET` is unset.
- Nothing in the logs distinguishes "missing config" from "Razorpay returned an error" or "network issue".
- Operators may spend time debugging payment flow or Razorpay integration before finding the real cause.

## Fix

Validate the secret at module load time so the process fails fast with a clear error message:

```ts
// At the top of payment.controller.ts, outside any handler
const RAZORPAY_KEY_SECRET = process.env.RAZORPAY_KEY_SECRET;
if (!RAZORPAY_KEY_SECRET) {
  throw new Error(
    "RAZORPAY_KEY_SECRET environment variable is required but not set. " +
    "Add it to your .env file before starting the server."
  );
}
```

Then use the validated constant inside the handler:

```ts
const expectedSignature = crypto
  .createHmac("sha256", RAZORPAY_KEY_SECRET)
  .update(body)
  .digest("hex");
```

This approach matches how `razorpay.ts` (the config file) should already guard `RAZORPAY_KEY_ID` and `RAZORPAY_KEY_SECRET`.

*GSSoC'26 contribution*

anshul23102:
## Summary

Fixes #1042

- Validates `RAZORPAY_KEY_SECRET` at **module load time** so a missing or empty environment variable throws a clear `Error` on server startup instead of silently passing `undefined` to `crypto.createHmac()` and returning an opaque `500` during a live payment verification request.
- Replaces the `expectedSignature !== razorpay_signature` string equality check with `crypto.timingSafeEqual()` to eliminate timing side-channel leakage when comparing HMAC signatures.
- The module-level constant is reused inside `verifyPayment`, removing the redundant inline `process.env.RAZORPAY_KEY_SECRET as string` cast.

## Changes

`backend/src/controllers/payment.controller.ts`
- Added module-level `RAZORPAY_KEY_SECRET` constant with fail-fast guard
- Updated `verifyPayment` to use the constant instead of re-reading `process.env`
- Switched signature comparison to `crypto.timingSafeEqual` with a length pre-check

## Test plan

- [ ] Start the server without `RAZORPAY_KEY_SECRET` set -- process should exit immediately with a clear error message
- [ ] Start the server with `RAZORPAY_KEY_SECRET` set -- server should boot normally
- [ ] Submit a valid payment -- `POST /payment/verify` returns `200 { success: true }`
- [ ] Submit a tampered signature -- `POST /payment/verify` returns `400 { success: false, message: "Invalid signature" }`

### DIFF
--- a/backend/src/controllers/payment.controller.ts
+++ b/backend/src/controllers/payment.controller.ts
@@ -2,6 +2,17 @@ import { Request, Response } from "express";
 import crypto from "crypto";
 import razorpayInstance from "../config/razorpay";
 
+// Validate RAZORPAY_KEY_SECRET is present at startup so misconfigured
+// deployments fail loudly rather than silently passing undefined to
+// crypto.createHmac() and returning an opaque 500 during payment verification.
+const RAZORPAY_KEY_SECRET = process.env.RAZORPAY_KEY_SECRET;
+if (!RAZORPAY_KEY_SECRET) {
+  throw new Error(
+    "Missing required environment variable: RAZORPAY_KEY_SECRET. " +
+    "Payment verification cannot work without it."
+  );
+}
+
 // Creates a new Razorpay order and returns the order details to the frontend
 export const createOrder = async (req: Request, res: Response) => {
   try {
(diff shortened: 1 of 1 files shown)


ronisarkarexe:
@anshul23102 Please give a star to support my project.

anshul23102:
> @anshul23102 Please give a star to support my project.

ofc.

anshul23102:
Could the maintainers add the `gssoc'26` label to this merged PR as well? It was filed under GirlScript Summer of Code 2026, and the tag keeps program tracking consistent across my contributions in this repo. The `gssoc:approved` and `level:beginner` labels are already in place. Thank you!

PR #1137, opened 2026-05-29
fix: move matplotlib evaluation code inside __main__ guard in train_scorer.py

### LINKED ISSUE #879: BUG: train_scorer.py runs matplotlib visualization code outside __main__ guard causing import side effects

### What happened?

At the bottom of backend/ml/train_scorer.py, the matplotlib evaluation and plotting code runs at module level instead of being placed inside the if __name__ == "__main__": block.

This means the following code executes automatically whenever the module is imported:

pred_coh, pred_cre, pred_rel = model.predict(...)
fig, axes = plt.subplots(1, 3, figsize=(15, 5))
...
plt.savefig("scorer_eval.png")

If another module imports from train_scorer.py, the evaluation code runs immediately and may call model.predict() before the model is initialized, causing crashes or runtime errors.

Expected behavior:

Evaluation and plotting code should only execute when the script is run directly.

Actual behavior:

Evaluation and plotting code executes on import.

### Steps to reproduce

Open backend/ml/train_scorer.py
Observe that the matplotlib evaluation code is outside the if __name__ == "__main__": block
Create another Python file and import the module:
from backend.ml.train_scorer import *
Run the importing file
Notice that the plotting/evaluation logic executes automatically during import
If the model is not initialized, model.predict() may fail

### Environment (if relevant)

Python 3.x
Matplotlib
TensorFlow / Keras model pipeline
Backend ML training module

N/A

### Record

shreyjeetdas:
## Screenshot (when helpful)
N/A — Python script fix, no UI involved.

## What this PR does
In `backend/ml/train_scorer.py`, the matplotlib evaluation and plotting 
code was running at module level outside the `if __name__ == "__main__":` 
block. This caused the evaluation code to execute automatically on import, 
calling `model.predict()` before the model was initialized and potentially 
crashing any module that imported from `train_scorer.py`.

This PR moves the matplotlib evaluation block inside the `if __name__ == 
"__main__":` guard so it only executes when the script is run directly.

## Type of change
- [x] Bug fix
- [ ] New feature
- [ ] Code refactor
- [ ] Documentation update

## Related issue
Fixes #879

N/A

## Checklist
- [ ] I have added screenshots or a recording when they help explain this PR, or noted N/A with a one-line reason.

### DIFF
--- a/backend/ml/train_scorer.py
+++ b/backend/ml/train_scorer.py
@@ -170,33 +170,33 @@ if __name__ == "__main__":
     print(f"Saved to {SAVE_DIR}/")
     
     
-import matplotlib.pyplot as plt
-import numpy as np
+    import matplotlib.pyplot as plt
+    import numpy as np
 
-# Predictions lao
-val_size = int(0.15 * len(stories))
-val_stories = story_enc[-val_size:]
-val_prompts = prompt_enc[-val_size:]
-val_coh, val_cre, val_rel = coh[-val_size:], cre[-val_size:], rel[-val_size:]
+    # Predictions lao
+    val_size = int(0.15 * len(stories))
+    val_stories = story_enc[-val_size:]
+    val_prompts = prompt_enc[-val_size:]
+    val_coh, val_cre, val_rel = coh[-val_size:], cre[-val_size:], rel[-val_size:]
 
-pred_coh, pred_cre, pred_rel = model.predict(
-    {"story": val_stories, "prompt": val_prompts}, verbose=0
-)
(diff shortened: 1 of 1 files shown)


shreyjeetdas:
The lint and typecheck CI failures are caused by package-lock.json being out of sync with package.json in the base branch — missing packages like @types/express, jest, razorpay, and ts-jest. This is a pre-existing issue unrelated to my changes, which only modify backend/ml/train_scorer.py (a Python file). The lock file needs to be updated in the base branch separately.

PR #1167, opened 2026-05-29
fix: add explicit auth middleware to GET /user/profile route

### LINKED ISSUE #770: Improve: GET /user/profile route missing explicit auth middleware

### What happened?

**File:** `backend/src/app/modules/user/user.router.ts` — line 14

The `GET /user/profile` route is the only protected route in the user router that does **not** have an explicit `auth(...)` middleware guard:

```ts
// user.router.ts

// All other protected routes have explicit auth middleware ✓
router.get("/lists", auth(ENUM_USER_ROLE.ADMIN, ENUM_USER_ROLE.SUPER_ADMIN), UserController.getAllUsers);
router.get("/:id", auth(ENUM_USER_ROLE.USER, ...), UserController.getUser);
router.patch("/update", auth(ENUM_USER_ROLE.USER, ...), UserController.updateUser);

// Profile route is missing auth middleware ✗
router.get("/profile", UserController.getProfileInfo);
```

The `getProfileInfo` controller calls `getToken(req)` internally, which throws a 401 if no token is present — so the route is not publicly accessible in practice. However, this is an implicit, undocumented behavior that creates a real risk:

- If `getProfileInfo` is ever refactored to not call `getToken` first, the route silently becomes public
- The inconsistency is confusing for contributors who expect all protected routes to be guarded at the router level
- It bypasses the standard auth middleware pattern used everywhere else in the codebase

### Steps to reproduce

1. Open `backend/src/app/modules/user/user.router.ts`
2. Compare the `/profile` route definition against every other route in the same file
3. Observe that `/profile` is the only route that handles auth implicitly inside the controller rather than explicitly at the router level

### Proposed fix

Add the `auth(...)` middleware to the profile route, consistent with the rest of the router:

```ts
router.get(
  "/profile",
  auth(
    ENUM_USER_ROLE.USER,
    ENUM_USER_ROLE.WRITER,
    ENUM_USER_ROLE.ADMIN,
    ENUM_USER_ROLE.SUPER_ADMIN
  ),
  UserController.getProfileInfo,
);
```

This is a one-line change that makes the security contract explicit and consistent with every other protected route in the project.

### Why does this matter?

- **Defense in depth:** explicit middleware is the established pattern; implicit controller-level auth is fragile
- **Contributor clarity:** new contributors reading the router can immediately see which routes are protected
- **Refactor safety:** future changes to `getProfileInfo` won't accidentally expose the endpoint

N/A — the issue is visible directly in the source file.

### Record

nancysangani:
**Issue:** Closes #770

---

## Screenshot (when helpful)

N/A — this is a backend middleware change with no UI impact.

---

## What this PR does

The `GET /user/profile` route in `backend/src/app/modules/user/user.router.ts` was the only protected route in the user router **without** an explicit `auth(...)` middleware guard.

Every other protected route in the same file guards access at the router level:

```ts
router.get("/lists", auth(ENUM_USER_ROLE.ADMIN, ENUM_USER_ROLE.SUPER_ADMIN), UserController.getAllUsers);
router.get("/:id",  auth(ENUM_USER_ROLE.USER, ...), UserController.getUser);
router.patch("/update", auth(ENUM_USER_ROLE.USER, ...), UserController.updateUser);
```

The `/profile` route was missing this and instead relied on an implicit 401 thrown inside `getProfileInfo` when `getToken(req)` finds no token. That works today, but is fragile — any refactor of `getProfileInfo` that moves or removes the `getToken` call would silently expose the endpoint.

This PR adds the explicit `auth(...)` guard to `/profile`, making it consistent with the rest of the router. **No behaviour changes** — unauthenticated requests still get a 401, authenticated requests still get the profile. The security contract is now just visible and enforced at the right layer.

---

## Type of change

- [ ] Bug fix
- [ ] New feature
- [x] Code refactor
- [ ] Documentation update

> Technically a security-hygiene / defensive refactor. The implicit auth still worked, but this makes the protection explicit and consistent with the rest of the codebase.

---

## Related issue

Closes #770

---

## Checklist

### DIFF
--- a/backend/src/app/modules/user/user.router.ts
+++ b/backend/src/app/modules/user/user.router.ts
@@ -11,7 +11,16 @@ const router = express.Router();
 router.get("/lists", auth(ENUM_USER_ROLE.ADMIN, ENUM_USER_ROLE.SUPER_ADMIN), UserController.getAllUsers);
 
 // Profile
-router.get("/profile", UserController.getProfileInfo);
+router.get(
+  "/profile",
+  auth(
+    ENUM_USER_ROLE.USER,
+    ENUM_USER_ROLE.WRITER,
+    ENUM_USER_ROLE.ADMIN,
+    ENUM_USER_ROLE.SUPER_ADMIN
+  ),
+  UserController.getProfileInfo,
+);
 
 // Apply for Writer
 router.get(


ronisarkarexe:

Issue #607, opened 2026-05-27
[Feat]: (GSSoC'26) Silent session expiry — access token expires without auto-refresh, causing silent API failures


Janarthanang-10:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### What should exist

The frontend should automatically refresh the expired access token using the existing refresh token endpoint, instead of silently failing on authenticated API calls.

### Who uses it

All logged-in users. Anyone whose session runs long enough for the access token to expire will hit this issue mid-session.

### What problem it solves

The backend already implements a full refresh token system:
- `POST /api/auth/refresh-token` endpoint exists and works
- `refreshToken` is set as an `httpOnly` cookie on login, register, and Google login

However, the frontend **never calls this endpoint**. When the access token expires, the Axios interceptor in `helpers/axios/axionInstance.ts` simply rejects the request with no retry logic:

```ts
// Current behaviour — no refresh token handling
instance.interceptors.response.use(
  (response) => response,
  function (error) {
    return Promise.reject(errorObject); // ❌ silently fails
  }
);
```

This means:
- Authenticated API calls silently fail after token expiry
- Users see broken UI or empty data with no explanation
- The refresh token cookie set by the backend is completely unused

### Proposed solution

Add token auto-refresh logic to the Axios response interceptor:

```ts
instance.interceptors.response.use(
  (response) => response,
  async (error) => {
    const originalRequest = error.config;
    if (error.response?.status === 401 && !originalRequest._retry) {
      originalRequest._retry = true;
      try {
        const { data } = await axios.post('/api/auth/refresh-token');
        const newToken = data.data.accessToken;
        localStorage.setItem('accessToken', newToken);
        originalRequest.headers.Authorization = newToken;
        return instance(originalRequest);
      } catch {
        localStorage.removeItem('accessToken');
        window.location.href = '/login';
      }
    }
    return Promise.reject(error);
  }
);
```

### Additional context

- No backend changes required — the refresh endpoint is already production-ready
- Fix is localised to `helpers/axios/axionInstance.ts`
- Secondary cleanup: remove the direct `localStorage.setItem("accessToken", token)` in `auth.context.tsx` and rely solely on `storeUserInfo()` to avoid dual-key inconsistency

### Acceptance criteria (optional)

- [ ] When an authenticated API call returns `401`, the Axios interceptor automatically calls `POST /api/auth/refresh-token`
[...]

PR #1299, opened 2026-05-29
feat: implement access token auto-refresh on 401 response

Janarthanang-10:
ISSUE NUMBER: #607

Implements access token auto-refresh using the existing `POST /api/auth/refresh-token` endpoint. Previously, when the access token expired, all authenticated API calls silently failed. Now they automatically retry with a refreshed token without any interruption to the user.

---

### `frontend/src/helpers/axios/axionInstance.ts`
- Added a 401 response interceptor that automatically calls `POST /api/auth/refresh-token` when an access token expires
- On success, stores the new token and retries the original request transparently
- On failure (refresh token also expired), clears local storage and redirects to `/login`
- Added `_retry` flag to prevent infinite retry loops
- Used `withCredentials: true` so the browser sends the `httpOnly` refresh token cookie set by the backend

### `frontend/src/components/auth.context.tsx`
- Removed duplicate `localStorage.setItem("accessToken", token)` call in the `login` function — `storeUserInfo()` is now the single source of truth for token storage
- Changed `localStorage.getItem("accessToken")` to use the `AUTH_KEY` constant to avoid hardcoded string divergence
- Simplified `logout` to use only `removeUserInfo()` instead of the redundant direct `localStorage.removeItem` call

---

## How to test

1. Login with any account
2. Open DevTools → Application → Local Storage → delete `accessToken`
3. Open Network tab → filter by Fetch/XHR
4. Trigger any authenticated action in the app
5. Observe `refresh-token` request fires automatically and the original request retries successfully
6. User stays logged in with no interruption ✅

---

## Notes

- No backend changes required — `POST /api/auth/refresh-token` was already production-ready
- Fix is localised to 2 files only

### DIFF
--- a/frontend/src/components/auth.context.tsx
+++ b/frontend/src/components/auth.context.tsx
@@ -1,6 +1,11 @@
 import React, { createContext, useState, useEffect, useCallback } from "react";
 import { useNavigate } from "react-router-dom";
-import { getUserInfo, removeUserInfo, storeUserInfo } from "../services/auth.service";
+import {
+  getUserInfo,
+  removeUserInfo,
+  storeUserInfo,
+} from "../services/auth.service";
+import { AUTH_KEY } from "../constants/storage-key";
 
 interface User {
   id: string;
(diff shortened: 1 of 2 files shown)


Janarthanang-10:
coulde please merge it ?

ronisarkarexe:

PR #1309, opened 2026-05-29
fix: rename errorMessage to errorMessages for consistency (#1291)

### LINKED ISSUE #1291: BUG: Backend sends "errorMessage" but the frontend reads "errorMessages" — detailed errors get lost

### What happened?

There's a naming mismatch in the error payload. The backend returns the field as errorMessage (singular) in both
backend/src/app/middleware/global.error.handler.ts and the 404 handler in backend/src/app.ts. But basically everything
   else expects errorMessages (plural): the backend's own IGenericErrorResponse interface, the frontend type in
  frontend/src/types/common.ts, and the RTK Query base query in frontend/src/helpers/axios/axios.base.query.ts, which
does err.errorMessages || ....

  The practical effect: when the API returns field-level validation errors, that frontend path reads errorMessages,
  finds nothing, and falls back to a generic message — so users never see the specific "this field is wrong" details.
  The two axios layers in the frontend also disagree with each other (axionInstance.ts reads the singular form), which
is a hint the naming drifted at some point.

### Steps to reproduce

1. Trigger a validation error from the API (e.g., submit a form with an invalid field).
2. Expected: the specific field error message shows up.
  3. Actual: a generic fallback message shows instead, because the structured field is named differently than the
frontend looks for.

  How I'd fix it
  Make the backend consistent with its own interface and the frontend by renaming the response field to "errorMessages."
  (plural) in global.error.handler.ts and app.ts. Then double-check axionInstance.ts so both frontend axios layers read
  the same plural field.

### Environment (if relevant)

Full stack (backend response shape + frontend error handling).

_No response_

### Record

yachikadev:
## What this PR does
Fixes a naming mismatch where backend was sending `errorMessage` (singular) in error responses, but the frontend and `IGenericErrorResponse` interface expected `errorMessages` (plural). This caused field-level validation errors to be silently dropped and users would only see a generic fallback message.

## Type of change
- [x] Bug fix

## Related issue
Fixes #1291

## Checklist

## Screenshot
N/A — backend naming fix, no UI change. The fix ensures validation error details reach the frontend correctly instead of falling back to a generic message.


ronisarkarexe:
Just a friendly reminder: since there are many open PRs, merge conflicts can sometimes happen when other changes are merged first.

To help avoid this, please keep your branch synced with the latest `main` branch and resolve any conflicts if they appear.

Thank you for your contribution and patience. I’m reviewing PRs as quickly as possible, and I truly appreciate your support and effort in improving the project. 🚀

yachikadev:
@ronisarkarexe 
CI failures are pre-existing and unrelated to this PR's change. The build was already failing on main before this fix.

yachikadev:
@ronisarkarexe "CI failures appear to be pre-existing and unrelated to this PR. Other PRs are also failing with the same checks."

PR #1404, opened 2026-05-30
docs: add missing Razorpay environment variables

Tharun-Varshan-S:
## Description
Added missing Razorpay environment variable placeholders to backend/.env.example.

## Changes
- Added RAZORPAY_KEY_ID
- Added RAZORPAY_KEY_SECRET

## Why
These variables are required by the backend configuration but were missing from .env.example, causing confusion during local setup.

### DIFF
--- a/backend/.env.example
+++ b/backend/.env.example
@@ -30,6 +30,10 @@ VERIFY_PASSWORD=
 # Google OAuth
 GOOGLE_CLIENT_ID=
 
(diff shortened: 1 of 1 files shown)


Tharun-Varshan-S (commented):
Tested locally by copying `backend/.env.example` to `.env` and verified the missing Razorpay placeholders are now included.

Issue #1392, opened 2026-05-30
bug: no load more button or pagination in the stories list


PrudenceShubh:
the list of stories does not have any load more button ot pagination concept 

if we add the fields then

this will imporve the up of the page

<img width="1153" height="832" alt="Image" src="https://github.com/user-attachments/assets/e9aaf40f-e74b-467e-9f5f-26785b09fb15" />

PrudenceShubh:
I'd like to work on this — contributing under GSSoC'26

AnushaRumi:
hi @ronisarkarexe can u plz assign this issue to me as it is unassigned

udaycodespace:
> hi [@ronisarkarexe](https://github.com/ronisarkarexe) can u plz assign this issue to me as it is unassigned

Hey I am too a GSSoC 26' Contributor myself 🙂‍↔️

PR #1406, opened 2026-05-30
 added the load more button in the latest post component

PrudenceShubh:
added oad more button in the latest post page so that it doesnot load at first and when the user click on the load more it will open all the other post initially it will load only 7 posts


PrudenceShubh:
solved this issue 

bug: no load more button or pagination in the stories list #1392

Issue #633, opened 2026-05-27
[Backend/Security] Fix Broken Error Middleware Cascade and Prevent Sensitive Data Leaks


Aryan0819:
**Problem Statement**
The configuration order of operations inside the global Express middleware stack is currently broken. The server's fallback 404 handler sits upstream from the globalErrorHandler and intercepts unmatched routes by sending a JSON payload directly without invoking next().

Because of this gap, any unexpected runtime or operational exceptions thrown inside active routers completely bypass the global error handling system. This can crash the Node process or leak raw, unformatted internal system trace logs directly to client payloads, exposing sensitive architecture information.

**Solution Path**
Reorder the Express pipeline execution chain so that the globalErrorHandler middleware sits as the absolute final declaration in the bootstrapping file.
Update the fallback 404 route capture handler to correctly format unmatched route data before passing errors downwards through the middleware chain using next().
Append strict payload memory constraints (limit: "2mb") on incoming JSON and URL-encoded body parsers to safeguard production processes against buffer overflow exploits and Denial of Service (DoS) attacks.

Please assign me under gssoc 2026

MansviJiwane:
Hi, I’m Mansvi. I’d like to work on this issue under GSSoc.Could you assign it to me?

MansviJiwane:
Yes, I am working on it and will complete it by tomorrow.

MansviJiwane:
Update: I have completed the implementation and submitted PR #1 for review. Waiting for maintainer feedback.

MansviJiwane:
Hi,

I had already completed the work and submitted PR #1 for this issue before the auto-unassignment happened. The issue was automatically unassigned due to inactivity, but my pull request is still open and linked to the issue.

Could you please review my PR when you get a chance?

Thank you.

PR #1509, opened 2026-05-30
fix(security): hide error stack traces outside development

### LINKED ISSUE #1508: BUG: Production API leaks full stack traces and internal file paths to anonymous callers (Critical)

### What happened?

Every error response from the live API includes a full Node.js `stack` trace, exposing absolute server file paths, the module layout, and dependency versions to any anonymous caller.

In `backend/src/app/middleware/global.error.handler.ts` the response is built as:

```
stack: config.env != "production" ? err.stack : undefined,
```

`config.env` is `process.env.NODE_ENV`. On the live deployment `NODE_ENV` is not `"production"`, so the condition is true and `err.stack` is returned to the client on every error. Because the check is fail-open (it leaks for everything except the exact string `"production"`), a single missing or misconfigured environment variable turns the leak on.

This is distinct from #633. #633 is about middleware ordering and the 404 handler not calling `next()`. This issue is about the `stack` field being serialized into the JSON response by the error handler itself, and the fail-open environment check that controls it. The leak happens even when the error handler runs exactly as intended.

What an attacker learns:
- Absolute server paths such as `/var/task/backend/src/app/modules/auth/auth.service.js`.
- Internal module structure and route-to-file mapping.
- Exact dependency versions, for example `/var/task/node_modules/.pnpm/mongoose@8.24.0/...`, which makes it easy to look up known CVEs for the pinned versions.

This maps to OWASP ASVS V7.4.1 (do not disclose internal error details) and CWE-209 (Generation of Error Message Containing Sensitive Information).

### Steps to reproduce

1. Hit any endpoint that produces an error. Example, an ApiError path:
   ```
   curl -sX POST https://apistorysparkai.vercel.app/api/v1/auth/forgot-password \
     -H "Content-Type: application/json" \
     -d '{"email":"nonexistent_zzzz_99999@example.com"}'
   ```
   Response includes:
   ```
   "stack":"Error: User not found!\n    at /var/task/backend/src/app/modules/auth/auth.service.js:163:15\n    at Generator.next ..."
   ```
2. Trigger a different error class to confirm it is not endpoint specific. Example, a CastError:
   ```
   curl -sX GET "https://apistorysparkai.vercel.app/api/v1/post/not-a-valid-objectid-xyz"
   ```
   Response includes:
   ```
   "stack":"CastError: Cast to ObjectId failed ... at /var/task/node_modules/.pnpm/mongoose@8.24.0/node_modules/mongoose/lib/schema/objectId.js:253:11 ..."
   ```
3. Both responses expose internal paths and dependency versions to an unauthenticated client.

### Environment (if relevant)

- API host: `apistorysparkai.vercel.app`
- File: `backend/src/app/middleware/global.error.handler.ts`
- Verified live on production at the time of filing.

### Suggested fix

Make the stack exposure fail-closed so it is only ever returned in explicit development:

```
stack: config.env === "development" ? err.stack : undefined,
```

With this, any value of `NODE_ENV` other than `development` (including unset, `production`, or `staging`) hides the stack. This closes the leak regardless of how the deployment environment is configured.

Note on `NODE_ENV` and Vercel: simply forcing `NODE_ENV=production` inside `backend/vercel.json` is not safe here. The Vercel build runs `npm run build` which calls `tsc`, and `typescript` is a devDependency. Setting `NODE_ENV=production` so that it applies during install would prune devDependencies and break the build. If correct production logging behavior is also wanted, `NODE_ENV=production` should be set as a runtime scoped Production environment variable in the Vercel dashboard, not in a way that affects the install phase. The code level fail-closed change above is the primary fix and does not depend on the environment variable.

Follow-up worth tracking separately: the final `else if (err instanceof Error)` branch still echoes the raw `err.message` for unexpected 500s. Unexpected server errors should return a generic message rather than the internal error text.

N/A. The `curl` responses above are sufficient proof.

### Record

Anexus5919:
## Screenshot (when helpful)

N/A. This is a backend security fix verified via `curl`. Reproduction and verification are described under What this PR does.

## What this PR does

Stops the live API from returning full Node.js stack traces to anonymous callers.

The global error handler serialized `err.stack` into every error response whenever `NODE_ENV` was not exactly `"production"`. Because the live deployment does not have `NODE_ENV=production`, this exposed absolute server paths (for example `/var/task/backend/src/app/modules/auth/auth.service.js`), the internal module layout, and pinned dependency versions (for example `mongoose@8.24.0`) to unauthenticated users. This is CWE-209 and OWASP ASVS V7.4.1.

The check was fail-open: it leaked for every environment except one exact string. This PR makes it fail-closed so the stack is only ever exposed in explicit development.

### Change

`backend/src/app/middleware/global.error.handler.ts`

```
- stack: config.env != "production" ? err.stack : undefined,
+ stack: config.env === "development" ? err.stack : undefined,
```

Now any value of `NODE_ENV` other than `development` (including unset, `production`, or `staging`) hides the stack. The leak is closed regardless of how the deployment environment is configured, so the fix does not depend on a separate env var change.

### Why this approach and not vercel.json

The linked issue suggested also forcing `NODE_ENV=production` in `backend/vercel.json`. I deliberately did not do that. The Vercel build runs `npm run build` which calls `tsc`, and `typescript` is a devDependency. Setting `NODE_ENV=production` so that it applies during install would prune devDependencies and break the build. If the maintainer also wants correct production logging behavior, `NODE_ENV=production` should be set as a runtime scoped Production environment variable in the Vercel dashboard instead. The code level fail-closed change here is the primary fix and stands on its own.

### How I verified

1. Reproduced the leak live before the change on two different error classes (an ApiError on `/auth/forgot-password` and a CastError on `/post/<bad id>`); both responses contained `stack` with internal paths and dependency versions.
2. Traced the new code path: in any non development environment `config.env === "development"` is false, so `stack` is `undefined` and is omitted from the JSON. Local development still keeps the stack when `NODE_ENV=development`.
3. Ran `tsc --noEmit` on the backend. The changed file compiles cleanly. The only pre-existing errors are in `story_version.service.ts` and are unrelated to this change.
4. Confirmed no test asserts on the `stack` field.

### Scope note

This PR is intentionally limited to the stack trace leak so it stays focused and easy to review. It does not change middleware ordering or the 404 handler (that is #633), and it does not change the `err.message` echo on unexpected 500s, which is worth a separate follow-up.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Code refactor
- [ ] Documentation update

## Related issue

Closes #1508

N/A.

## Checklist

### DIFF
--- a/backend/src/app/middleware/global.error.handler.ts
+++ b/backend/src/app/middleware/global.error.handler.ts
@@ -73,7 +73,9 @@ const globalErrorHandler: ErrorRequestHandler = (
     success: false,
     message,
     errorMessage,
-    stack: config.env != "production" ? err.stack : undefined,
+    // Expose stack only in explicit development; hide by default so any
+    // non-development environment (including an unset NODE_ENV) stays safe.
+    stack: config.env === "development" ? err.stack : undefined,
   });
 };
 


Anexus5919:
@ronisarkarexe  Kindly have a review on this pr. Thanks!

Issue #1220, opened 2026-05-29
BUG: Frontend and Backend Applications Fail to Compile [CRITICAL]


mohammed-danyal:
### Is there an existing issue for this?

- [x] I have searched the existing issues

## Bug Description

The project currently cannot be built successfully as both the frontend and backend applications fail during the build process.

This issue blocks contributors from verifying changes locally and prevents successful deployment of the application.

---

## Problem

Running the build commands for the frontend and backend results in compilation/build errors that stop the process before completion.

This impacts:

* Local development
* CI/CD pipelines
* Production deployments
* Contributor onboarding and testing

---

## Expected Behavior

Both frontend and backend should compile successfully without errors.

```text
✓ Frontend build completed
✓ Backend build completed
```

---

## Actual Behavior

The build process fails due to compilation and/or type-checking errors in both frontend and backend modules.

---

## Acceptance Criteria

* [ ] Frontend builds successfully
* [ ] Backend builds successfully
* [ ] No TypeScript compilation errors remain
* [ ] No build-time errors remain
* [ ] Existing functionality remains unaffected
* [ ] `npm run build` completes successfully

---

## Environment

* Node.js: 20.x
* Frontend: React + TypeScript + Vite
* Backend: Node.js + Express + TypeScript

---

## Additional Context

This is currently a blocking issue affecting local development, testing, and deployment workflows.

Resolving all build failures will improve contributor experience, restore build stability, and ensure the application can be deployed successfully.

### Steps to reproduce

# Reproduction Steps

1. Clone the repository.

```bash
git clone https://github.com/ronisarkarexe/story-spark-ai.git
cd story-spark-ai
```

2. Install dependencies.

```bash
npm install
```

3. Attempt to build the project.

```bash
npm run build
```

Or build the applications individually:

```bash
npm run build:frontend
npm run build:backend
```

---

### Record

- [x] I agree to follow this project's Code of Conduct
- [ ] I want to work on this issue

akshayad2006-cmd:
Hi @ronisarkarexe 
I'd like to work on this — contributing under GSSoC'26. Could you please assign this issue to me? /assign

akshayad2006-cmd:
@ronisarkarexe I have fixed the bug and can you please check the pr

Issue #1161, opened 2026-05-29
BUG: npm typecheck CI fails because package-lock is out of sync


DhruvalBhinsara1:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### What happened?

The Type Check workflow currently fails during dependency installation before backend or frontend type checking can run because `package-lock.json` is out of sync with the current package manifests.

CI logs show:

```text
npm error `npm ci` can only install packages when your package.json and package-lock.json or npm-shrinkwrap.json are in sync.
npm error Missing: @types/express@4.17.25 from lock file
npm error Missing: jest@29.7.0 from lock file
npm error Missing: razorpay@2.9.6 from lock file
npm error Missing: ts-jest@29.4.11 from lock file
```

This affects unrelated pull requests because the workflow exits at `npm ci --ignore-scripts` before it can validate the actual PR changes.

Note: #1074 / #1086 cover the separate `pnpm-lock.yaml` drift in the lint workflow. This issue is for the npm lockfile used by `.github/workflows/typecheck.yml`.

### Steps to reproduce

1. Start from the current `main` branch.
2. Run the Type Check workflow install command:

```bash
npm ci --ignore-scripts
```

3. Observe that npm exits with `EUSAGE` because `package-lock.json` is stale.

### Environment (if relevant)

- GitHub Actions Type Check workflow
- Node 20 in CI
- npm clean install command: `npm ci --ignore-scripts`
- Root npm workspaces monorepo

### Screenshots or screen recording (if helpful)

N/A; the CI log above is the relevant evidence.

### Record

- [x] I agree to follow this project's Code of Conduct
- [x] I want to work on this issue

PR #1517, opened 2026-05-30
fix(payment): upgrade subscription on verified payment with persisted orders

### LINKED ISSUE #1516: BUG: Verified Razorpay payments never upgrade the subscription, and amount is client controlled (Critical)

### What happened?

The payment flow looks complete but a paid user is never actually upgraded, and the price is trusted from the client.

In `backend/src/controllers/payment.controller.ts`:

1. `verifyPayment` validates the Razorpay HMAC signature correctly and returns `{ success: true, message: "Payment verified successfully" }`, but it never updates the user's `subscriptionType`. No code anywhere in the backend writes `subscriptionType` after user creation, so a user can complete a real Razorpay payment and stay on the free tier.
2. There is no order record. `createOrder` does not persist anything linking `userId` to `razorpay_order_id` and the chosen plan. So even if you wanted to upgrade on verify, the server has no way to know which tier was paid for: the frontend `verify` call only forwards `razorpay_order_id`, `razorpay_payment_id`, and `razorpay_signature`, not the plan.
3. The amount is client controlled. `createOrder` reads `amount` from the request body and creates the Razorpay order with it, so a client can pay any amount they choose for any plan.

This is distinct from #1041, #1042, #1043, and #1298, which are about authentication and rate limiting on the payment routes. None of them address the subscription never being upgraded, the missing order record, or the client controlled amount.

This maps to CWE-840 (Business Logic Errors) and OWASP API6:2023 (Unrestricted Access to Sensitive Business Flows).

### Steps to reproduce

1. Call `POST /api/v1/payment/create-order` while logged in with a body like `{ "amount": 1 }`. A Razorpay order is created for that exact amount, proving the price is not enforced server side.
2. Complete the signature handshake and call `POST /api/v1/payment/verify`. It returns success.
3. Query the user document or decode a fresh access token after login. `subscriptionType` is still `free`, so no upgrade happened.

### Environment (if relevant)

- API host: `apistorysparkai.vercel.app`
- Files: `backend/src/controllers/payment.controller.ts`, `backend/src/router/payment.route.ts`

### Suggested fix

1. Persist orders. On `create-order`, store `{ userId, razorpayOrderId, plan, amount, currency, status: "created" }`.
2. Resolve the price server side from a plan to price map. Take a `plan` from the client, not an `amount`, so the user cannot set their own price.
3. On `verify`, after the signature passes, atomically claim the order (`status: created -> paid`) and update `User.subscriptionType` to the order's tier. The atomic status transition makes a replayed verify a no-op so a tier cannot be granted twice.
4. As a follow-up, add a Razorpay webhook handler with signature verification so the upgrade still happens if the browser closes before the verify call.

### Additional note (separate frontend bug)

While verifying this, the payment page `frontend/src/components/home/pricing/payment.component.tsx` appears broken on its own: the Razorpay handler `handlePayment` is never called (the form `onSubmit` calls an undefined `handlePay`), and several referenced state values (`name`, `cardNumber`, `cvv`, `loading`, `isFormValid`) are never declared. It also collects raw card fields in a custom form, which Razorpay should handle in its own checkout. This needs a separate frontend fix so the page actually calls `create-order` with a `plan`. The backend fix is correct and ready for whenever that frontend work lands.

A PR implementing items 1 to 3 on the backend is attached.

N/A. Reproduction is via the API calls and decoding the token as described above.

### Record

Anexus5919:
## Screenshot (when helpful)

N/A. Backend billing logic. Verified via type checks and a code trace, described below.

## What this PR does

Makes a verified Razorpay payment actually upgrade the user, records every order, and stops trusting the price from the client.

### The bugs

1. `verifyPayment` validated the signature and returned success but never set `subscriptionType`, so paying users stayed on free.
2. No order was persisted, so the server had no way to map a payment back to a user and a tier (the verify call only carries the Razorpay ids, not the plan).
3. `createOrder` trusted a client supplied `amount`, so a user could pay any price for a plan.

### Backend changes

New payment module under `backend/src/app/modules/payment/`:
- `order.model.ts` and `order.interface.ts`: an `Order` collection with `userId`, unique `razorpayOrderId`, `plan`, `amount`, `currency`, and `status` (`created` / `paid` / `failed`).
- `payment.constant.ts`: `PLAN_PRICING`, the server side source of truth for plan prices in paise (INR), plus a `normalizePlan` helper that maps accepted labels to the paid plan keys (`pro`, `premium`, with `enterprise` treated as `premium`).

`controllers/payment.controller.ts`:
- `createOrder` now takes a `plan`, resolves the price from `PLAN_PRICING` (the client can no longer set the amount), creates the Razorpay order, and persists the order against the authenticated user with `status: created`.
- `verifyPayment` verifies the signature, then atomically claims the order with a `created -> paid` transition scoped to that user, and upgrades `User.subscriptionType` to the order's tier. The atomic transition makes a replayed verify a no-op, so a tier cannot be granted twice. An already paid order owned by the same user returns success idempotently.

### Effect timing

Quota enforcement reads `subscriptionType` from the database (`quota.service.ts` and `ai_model.service.ts` load the user document), so the higher limit applies immediately after upgrade. The plan label embedded in the JWT refreshes on next login.

### Why this is backend only

The payment page `frontend/src/components/home/pricing/payment.component.tsx` is broken independently of this change: its Razorpay handler `handlePayment` is never called (the form `onSubmit` calls an undefined `handlePay`), and the state it references (`name`, `cardNumber`, `cvv`, `loading`, `isFormValid`) is never declared. Wiring it to send a `plan` is not a small change, it needs the component repaired first, so I kept that out of this PR to keep the security fix focused and reviewable. Once the page is fixed to call `create-order`, it should send `{ plan: "pro" }` or `{ plan: "premium" }` instead of an `amount`. Tracked in the linked issue.

### How I verified

1. Confirmed the original `verifyPayment` never wrote `subscriptionType` and that no other code path does either.
2. Traced the new flow: create-order persists `{ userId, razorpayOrderId, plan, amount }`; verify claims the order atomically and upgrades the user.
3. Ran `tsc` on the backend. The new payment files and the controller compile cleanly. The only remaining backend errors are the pre-existing ones in `story_version.service.ts`, unrelated to this PR.

### Out of scope (follow-ups)

- A Razorpay webhook handler as a server to server fallback.
- Repairing the broken payment page so it calls `create-order` with a `plan`.
- The pricing UI naming and currency mismatch (Starter/Pro/Enterprise and mixed $/₹ versus the backend free/pro/premium and INR).

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Code refactor
- [ ] Documentation update

## Related issue

Closes #1516

N/A.

## Checklist


Anexus5919:

Anexus5919:
@ronisarkarexe A quick note on the three red checks (CI/build, Lint, Type Check): they originate from the base branch, not from this change.

- Type Check and CI install fail at the `npm ci` / install step because the committed lockfile is out of sync with `package.json` (for example `helmet`, `express-rate-limit`, `ioredis`, and a `node-cron` version bump are missing from the lock file). This is already tracked in #1161 and #1534.
- CI build also hits pre-existing TypeScript errors in files this PR does not touch, for example `src/app.ts` and `recommendation.service.ts`. These are covered by #1220, #1544, and #1546.
- Lint reports unused-variable and similar errors in components this PR does not touch.

This PR only adds the payment module files and updates `payment.controller.ts`. The same three checks are red on `main` itself, so they are not caused by this change, and "No conflicts with base branch" is green. The merge from upstream has already been done. Happy to rebase once the base-branch CI issues above are resolved.

PR #1520, opened 2026-05-30
fix(security): back rate limiters with MongoDB so limits hold on serverless

### LINKED ISSUE #1519: BUG: In-memory rate limiters do not enforce on Vercel serverless (auth and free-AI limits bypassable)

### What happened?

The authentication and free-AI rate limiters keep their state in a per-process in-memory `Map`, so they do not work on the Vercel serverless deployment.

- `backend/src/app/middleware/ip.rate-limiter.ts` uses `const store = new Map(...)` for the login, registration, forgot-password, and reset-password limiters.
- `backend/src/app/middleware/free-ai.rate-limiter.ts` uses the same in-memory `Map` for guest AI generations.

On Vercel, each serverless function instance has its own memory, and instances are created and destroyed on demand. Requests are spread across many instances, and a cold start begins with an empty `Map`. So the counters are never shared. An attacker can bypass the limits simply by sending requests that land on different or fresh instances. This defeats the protection these limiters were added for: credential stuffing on login, registration and OTP spam, and free-AI abuse.

This is different from #1290, which is about a memory leak (missing cleanup) in `middlewares/rateLimitMiddleware.ts`. That issue assumes a single long-lived process. This issue is that the in-memory approach itself cannot enforce a shared limit on serverless.

This maps to CWE-307 (Improper Restriction of Excessive Authentication Attempts) and OWASP API4:2023 (Unrestricted Resource Consumption).

### Steps to reproduce

1. Send several requests in quick succession to a limited endpoint, for example forgot-password which is configured for 3 per hour:
   ```
   for i in $(seq 1 8); do
     curl -s -o /dev/null -w "%{http_code}\n" -X POST \
       https://apistorysparkai.vercel.app/api/v1/auth/forgot-password \
       -H "Content-Type: application/json" \
       -d '{"email":"probe_zzz_99999@example.com"}'
   done
   ```
2. Observe that none of the responses are `429`. With a working 3-per-hour limit, requests after the third should be blocked. I ran this against production and saw eight `404` responses and zero `429`.

### Environment (if relevant)

- API host: `apistorysparkai.vercel.app` (Vercel serverless, see `backend/vercel.json`)
- Files: `backend/src/app/middleware/ip.rate-limiter.ts`, `backend/src/app/middleware/free-ai.rate-limiter.ts`

### Suggested fix

Move the limiter state to a shared store so it holds across all instances. The conventional options are a MongoDB collection with a TTL index, or a managed Redis such as Upstash. Since this project already requires MongoDB and does not have a `REDIS_URL` provisioned (the existing `redis.client.ts` falls back to `127.0.0.1:6379`, which does not exist on Vercel), MongoDB is the lower-risk choice and needs no new infrastructure.

The attached PR adds a shared MongoDB-backed store with a TTL index and rewires both limiters to use it, preserving the exact thresholds and block windows, and failing open if the store is briefly unavailable so a database blip cannot lock users out of auth endpoints.

Related, out of scope here: the global `express-rate-limit` in `app.ts` uses the library's default in-memory store and has the same serverless limitation; and `app.ts` references an unimported `storyRoutes`. Worth separate issues.

N/A. The `curl` loop output above is the proof.

### Record

Anexus5919:
## Screenshot (when helpful)

N/A. Backend middleware change. Verified live (the bug) and via type checks and a code trace, described below.

## What this PR does

Makes the authentication and free-AI rate limiters actually enforce their limits on the Vercel serverless deployment.

### The bug

`ip.rate-limiter.ts` (login, registration, forgot-password, reset-password) and `free-ai.rate-limiter.ts` stored their counters in a per-process in-memory `Map`. On Vercel each function instance has its own memory and cold starts begin empty, so the counters are never shared and the limits are bypassable. I confirmed this live: eight forgot-password requests (limit 3 per hour) all returned `404` and none returned `429`.

### The fix

- New `backend/src/app/middleware/rate_limit.store.ts`: a shared MongoDB-backed store (`rate_limit_records`) with a TTL index on `expireAt`, plus `consumeRateLimit(...)`. The window-and-block state transition runs as a single atomic aggregation-pipeline `findOneAndUpdate` (the same atomic pattern already used in `quota.service.ts`), so concurrent requests cannot race. It fails open and logs if the store is briefly unavailable, so a database blip cannot lock everyone out of auth.
- `ip.rate-limiter.ts`: `createRateLimiter` now awaits the shared store instead of a `Map`. All four pre-configured limiters keep their exact thresholds and block windows (registration 5/hour with 24h block, login 10/15min, forgot 3/hour, reset 5/hour). Added an optional `buildMessage` so callers can keep their own 429 wording.
- `free-ai.rate-limiter.ts`: now defined via the same factory (5 per 24h, 24h block), preserving its original message. The duplicated in-memory logic and the in-process cleanup interval are gone (the TTL index handles expiry).

No route or call site changes: the exported limiters are still Express middleware (now async, which Express handles natively). Thresholds are unchanged.

### Why MongoDB and not Redis

Mongo is already required and connected. There is no `REDIS_URL` provisioned and `redis.client.ts` defaults to `127.0.0.1:6379`, which does not exist on Vercel, so a Redis-backed limiter would fail there. Mongo needs no new infrastructure or env var. (Confirmed with the maintainer direction to use Mongo and fail open.)

### How I verified

1. Reproduced the bypass live before the change (eight requests, zero 429).
2. Traced the new atomic pipeline for the new-key, within-window, window-elapsed, just-exceeded, and already-blocked cases; the post-update `blockedUntil > now` check yields the correct allow/deny in each, matching the previous behavior.
3. Ran `tsc` on the backend. The three changed files compile cleanly. The only remaining errors are the pre-existing ones in `story_version.service.ts`, unrelated to this PR.
4. Confirmed no test references these limiters (the existing rate-limit test targets the separate `middlewares/rateLimitMiddleware.ts`).

### Out of scope (follow-ups)

- The global `express-rate-limit` in `app.ts` uses the library default in-memory store and has the same serverless limitation.
- `app.ts` references an unimported `storyRoutes`.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Code refactor
- [ ] Documentation update

## Related issue

Closes #1519

N/A.

## Checklist

### DIFF
--- a/backend/src/app/middleware/free-ai.rate-limiter.ts
+++ b/backend/src/app/middleware/free-ai.rate-limiter.ts
@@ -1,87 +1,18 @@
-import { Request, Response, NextFunction } from "express";
-import ApiError from "../../errors/api_error";
-import httpStatus from "http-status";
+import { createRateLimiter } from "./ip.rate-limiter";
 
-interface RateRecord {
-  count: number;
-  firstRequestAt: number;
-  blockedUntil?: number;
-}
-
-const store = new Map<string, RateRecord>();
-
-// Configurable limits
-const WINDOW_MS = 24 * 60 * 60 * 1000; // 24 hours window
-const MAX_REQUESTS = 5; // max free generations per WINDOW_MS
-const BLOCK_TIME_MS = 24 * 60 * 60 * 1000; // block 24 hours when exceeded
-
-// Cleanup old keys periodically to prevent memory leak
-const cleanupInterval = setInterval(() => {
-  const now = Date.now();
-  for (const [key, rec] of store.entries()) {
-    const age = now - rec.firstRequestAt;
-    if ((rec.blockedUntil && rec.blockedUntil < now) || age > WINDOW_MS + BLOCK_TIME_MS) {
-      store.delete(key);
-    }
-  }
-}, 60 * 60 * 1000); // every hour
-cleanupInterval.unref(); // Allow Node process to exit cleanly
-
-export const freeAiRateLimiter = (req: Request, res: Response, next: NextFunction) => {
-  try {
-    const ip = req.ip;
-    
-    if (!ip) {
-      throw new ApiError(httpStatus.FORBIDDEN, "Could not determine client IP address.");
-    }
-    
-    const now = Date.now();
-    const key = `free_ai_${ip}`;
-
-    let rec = store.get(key);
-    if (!rec) {
-      rec = { count: 1, firstRequestAt: now };
-      store.set(key, rec);
-      return next();
-    }
-
-    // If currently blocked
-    if (rec.blockedUntil && rec.blockedUntil > now) {
-      const retryAfter = Math.ceil((rec.blockedUntil - now) / 1000);
-      res.setHeader("Retry-After", String(retryAfter));
-      throw new ApiError(
-        httpStatus.TOO_MANY_REQUESTS,
-        `Daily limit for free generations reached. Try again after ${Math.ceil(retryAfter / 3600)} hours or sign in to use your monthly quota.`
-      );
-    }
-
-    // Reset window if elapsed
-    if (now - rec.firstRequestAt > WINDOW_MS) {
-      rec.count = 1;
-      rec.firstRequestAt = now;
-      rec.blockedUntil = undefined;
-      store.set(key, rec);
-      return next();
-    }
-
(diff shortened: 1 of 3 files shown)


Anexus5919:
@ronisarkarexe Kindly have a review on this pr. Thanks!

Issue #1275, opened 2026-05-29
Guest user ID generated with Math.random()


itz-deeptanshu:
In ai_model.controller.ts, unauthenticated users get a cookie based ID created with Math.random().toString(36).substring(7). This produces only 5 characters — roughly 60 billion collisions possible among users. Guest quota can be easily bypassed by clearing the cookie. Use crypto.randomUUID() or uuid v4.

soniya28gupta:
Hi, I'd like to work on this issue under GSSoC'26. Please assign it to me.

MajBlaze:
Hi, I'd like to work on this issue under GSSoC'26. 
Please assign it to me.

soniya28gupta:
I am doing work and solve the issue in this repo

ManishaGurugubelli:
/assign

I'd like to work on this issue under GSSoC'26.

PR #1522, opened 2026-05-30
fix(security): generate OTP with crypto.randomInt instead of Math.random

### LINKED ISSUE #1521: BUG: Email/password-reset OTP generated with Math.random() is predictable

### What happened?

The email verification and password reset OTP is generated with `Math.random()`, which is not cryptographically secure.

`backend/src/app/modules/verify_email/verify_email.service.ts`:

```
const otp = Math.floor(100000 + Math.random() * 900000).toString();
```

`Math.random()` is backed by V8's xorshift128+ PRNG, which is seeded once per process and is not designed to be unpredictable. An attacker who observes a few legitimate outputs from the same Node process can recover the internal PRNG state and predict subsequent values. On a serverless cold start the available entropy is also low. Because this same OTP is used both for signup verification and for the password reset flow, a predictable OTP weakens account takeover protection.

This is different from #946, which was about rate limiting the OTP send endpoints, not the quality of the random number. It is also different from #1275, which is about the guest user id using `Math.random()` in a different file.

This maps to CWE-338 (Use of Cryptographically Weak PRNG).

### Steps to reproduce

1. Read `backend/src/app/modules/verify_email/verify_email.service.ts`.
2. The OTP is `Math.floor(100000 + Math.random() * 900000)`, which derives a security-sensitive code from a non-cryptographic PRNG.

### Environment (if relevant)

- File: `backend/src/app/modules/verify_email/verify_email.service.ts`
- Affects both the signup email verification and the forgot/reset password OTP flows.

### Suggested fix

Use Node's cryptographically secure RNG. `crypto` is already imported in this file:

```
const otp = crypto.randomInt(100000, 1000000).toString();
```

`crypto.randomInt(100000, 1000000)` returns an integer in `[100000, 1000000)`, i.e. 100000 to 999999, which is the same 6-digit range as before, so nothing downstream changes.

A one-line PR is attached.

N/A. This is a code-level cryptographic weakness.

### Record

Anexus5919:
## Screenshot (when helpful)

N/A. One-line backend cryptographic fix. Verified via type check and range equivalence, described below.

## What this PR does

Replaces the OTP generator with a cryptographically secure RNG.

Before, the email verification and password reset OTP was derived from `Math.random()`:

```
const otp = Math.floor(100000 + Math.random() * 900000).toString();
```

`Math.random()` is not cryptographically secure (V8 xorshift128+, seeded once per process), so its output is predictable from a few observed samples, and entropy is low on serverless cold starts. This OTP gates both signup verification and password reset, so predictability weakens account takeover protection. This is CWE-338.

After:

```
const otp = crypto.randomInt(100000, 1000000).toString();
```

`crypto` was already imported in this file (it is used for the verification tokens). `crypto.randomInt(100000, 1000000)` returns an integer in `[100000, 1000000)`, i.e. 100000 to 999999, the exact same 6-digit range as before, so the OTP format and everything downstream is unchanged.

## How I verified

1. Confirmed this is the only OTP generation site; the other `Math.random()` uses are the guest user id (tracked separately in #1275) and the collab room id, which this PR does not touch.
2. Confirmed `crypto.randomInt(100000, 1000000)` yields a 6-digit string in the same range as the old expression.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Code refactor
- [ ] Documentation update

## Related issue

Closes #1521

N/A.

## Checklist

### DIFF
--- a/backend/src/app/modules/verify_email/verify_email.service.ts
+++ b/backend/src/app/modules/verify_email/verify_email.service.ts
@@ -18,7 +18,8 @@ const transporter = nodemailer.createTransport({
 const VerifyEmail = async (payload: IEmailBody) => {
   try {
     const { email, name } = payload;
-    const otp = Math.floor(100000 + Math.random() * 900000).toString();
+    // Use a cryptographically secure RNG so OTPs cannot be predicted.
+    const otp = crypto.randomInt(100000, 1000000).toString();
     const expiresAt = new Date(Date.now() + 10 * 60 * 1000); // 10 minutes from now
     
     // Delete any existing OTP for this email


Anexus5919:
@ronisarkarexe Please review this pr whenever you get time. Thanks!

PR #1524, opened 2026-05-30
fix(security): pin JWT algorithm to HS256 on sign and verify

### LINKED ISSUE #1523: BUG: JWT verification does not pin the signing algorithm

### What happened?

The shared JWT verification helper does not pin an algorithm, so it accepts whatever the token's `alg` header claims.

`backend/src/utils/jwt.helper.ts`:

```
const verifyToken = (token: string, secret: Secret): JwtPayload => {
  return jwt.verify(token, secret) as JwtPayload;
};
```

Every token check in the app routes through this one helper (`auth.middleware.ts`, `check.request.limit.ts`, `token.ts`, the refresh-token path in `auth.service.ts`, the Socket.IO handshake in `server.ts`, and the collab namespace in `collab.socket.ts`), so the missing pin affects all of them.

Tokens are signed with HS256 (`createToken` uses the default, `createResetToken` sets HS256 explicitly). Without an `algorithms` allowlist on verify, the verifier honors the attacker-controlled `alg` header. `jsonwebtoken` v9 closed the `none` loophole, but not pinning still leaves the door open to algorithm-confusion and downgrade issues if an asymmetric key path is ever introduced, and it is simply not defense in depth. Pinning the algorithm is the standard hardening step and costs one line.

This maps to CWE-347 (Improper Verification of Cryptographic Signature).

### Steps to reproduce

1. Read `backend/src/utils/jwt.helper.ts`. `verifyToken` calls `jwt.verify(token, secret)` with no `algorithms` option.
2. The verifier therefore does not restrict which algorithm a presented token may use.

### Environment (if relevant)

- File: `backend/src/utils/jwt.helper.ts`
- Affects all auth-protected REST routes and both Socket.IO namespaces, since they all use this helper.

### Suggested fix

Pin the algorithm on verify, and set it explicitly on sign so the two never drift:

```
// sign
const options = { algorithm: "HS256", expiresIn: expireTime } as SignOptions;

// verify
return jwt.verify(token, secret, { algorithms: ["HS256"] }) as JwtPayload;
```

Because every call site delegates to this helper, this single change covers all of them. A one-line-per-function PR is attached.

N/A. This is a code-level cryptographic hardening change.

### Record

Anexus5919:
## Screenshot (when helpful)

N/A. Backend cryptographic hardening. Verified via type check and a call-site trace, described below.

## What this PR does

Pins the JWT algorithm so the verifier no longer trusts the token's `alg` header.

`backend/src/utils/jwt.helper.ts` `verifyToken` called `jwt.verify(token, secret)` with no `algorithms` allowlist, so it accepted whatever algorithm a presented token claimed. Tokens are signed with HS256. This is CWE-347 and standard hardening that was missing.

Changes:
- `verifyToken` now passes `{ algorithms: ["HS256"] }`.
- `createToken` now sets `algorithm: "HS256"` explicitly (it already defaulted to HS256; this makes sign and verify explicitly aligned and matches the existing `createResetToken`).

### Why this one file covers everything

Every token check in the app delegates to `JwtHalers.verifyToken`: `auth.middleware.ts`, `check.request.limit.ts`, `token.ts`, the refresh path in `auth.service.ts`, the Socket.IO handshake in `server.ts`, and the collab namespace in `collab.socket.ts`. Pinning the algorithm in the shared helper hardens all of them at once, so no call site needs to change.

## How I verified

1. Confirmed all six verify call sites route through `JwtHalers.verifyToken`, so the central change applies everywhere.
2. Confirmed tokens are signed with HS256 (default in `createToken`, explicit in `createResetToken`), so `["HS256"]` matches and existing tokens keep verifying.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Code refactor
- [ ] Documentation update

## Related issue

Closes #1523

N/A.

## Checklist

### DIFF
--- a/backend/src/utils/jwt.helper.ts
+++ b/backend/src/utils/jwt.helper.ts
@@ -5,7 +5,7 @@ const createToken = (
   secret: Secret,
   expireTime: string
 ): string => {
-  const options = { expiresIn: expireTime } as SignOptions;
+  const options = { algorithm: "HS256", expiresIn: expireTime } as SignOptions;
   return jwt.sign(payload, secret, options);
 };
 
@@ -22,7 +22,8 @@ const createResetToken = (
 };
 
 const verifyToken = (token: string, secret: Secret): JwtPayload => {
-  return jwt.verify(token, secret) as JwtPayload;
+  // Pin the algorithm so a forged token cannot downgrade or switch the alg header.
+  return jwt.verify(token, secret, { algorithms: ["HS256"] }) as JwtPayload;
 };
 
 export const JwtHalers = {


Anexus5919:
Thanks!

ronisarkarexe:

Anexus5919:
> Please give a star to support my project.

@ronisarkarexe Sure.

PR #1528, opened 2026-05-30
fix(security): reject Google login when email is not verified

### LINKED ISSUE #1527: BUG: Google OAuth login does not check email_verified (account takeover vector)

### What happened?

The Google OAuth login verifies the ID token signature but never checks the `email_verified` claim before trusting the email.

`backend/src/app/modules/auth/auth.service.ts`, in `googleLogin`:

```
const payload_data = ticket.getPayload();
if (!payload_data || !payload_data.email) {
  throw new ApiError(httpStatus.BAD_REQUEST, "Invalid Google token");
}

const { email, name: googleName, picture } = payload_data;
let user = await User.findOne({ email });
// ... creates / logs in the user keyed only by email
```

Google's OIDC payload includes `email_verified` precisely because an email present in a token is not proof of ownership. With a Google Workspace, an admin can create accounts with arbitrary, unverified email addresses. If `email_verified` is not checked, someone could obtain a validly signed Google ID token carrying an email they do not own and have this backend create or log into an account under that email. Since accounts here are keyed by email, that is an account takeover and impersonation vector.

This maps to CWE-287 (Improper Authentication).

### Steps to reproduce

1. Read `backend/src/app/modules/auth/auth.service.ts` `googleLogin`. After `verifyIdToken`, only `payload_data.email` is checked, not `payload_data.email_verified`.
2. In a test setup with your own client id, present a validly signed Google ID token whose payload has `email_verified: false`. The server still creates or logs in the user.

### Environment (if relevant)

- File: `backend/src/app/modules/auth/auth.service.ts` (`googleLogin`)
- Endpoint: `POST /api/v1/auth/google-login`

### Suggested fix

Reject the login when the email is not verified, right after the existing payload check:

```
if (!payload_data.email_verified) {
  throw new ApiError(httpStatus.UNAUTHORIZED, "Google email is not verified");
}
```

`google-auth-library` exposes `email_verified` on the token payload, so this is a small, well-typed guard. A PR is attached.

N/A. This is a code-level authentication weakness.

### Record

Anexus5919:
## Screenshot (when helpful)

N/A. Backend authentication hardening. Verified via type check and a code trace, described below.

## What this PR does

Adds the missing `email_verified` check to Google OAuth login.

`googleLogin` in `backend/src/app/modules/auth/auth.service.ts` verified the ID token signature and that an `email` was present, but never checked `payload_data.email_verified`. Because accounts are keyed by email, a validly signed Google token carrying an unverified email the caller does not own could create or log into an account under that email. This is an account takeover and impersonation vector (CWE-287).

Change, added right after the existing payload check:

```
if (!payload_data.email_verified) {
  throw new ApiError(httpStatus.UNAUTHORIZED, "Google email is not verified");
}
```

This blocks both new-user creation and login for any Google token whose email is not verified, while normal Gmail and Workspace accounts (which are verified) are unaffected.

## How I verified

1. Confirmed `googleLogin` only checked `payload_data.email`, not `email_verified`, before creating or logging in the user.
2. `google-auth-library` exposes `email_verified` on the token payload, and `!payload_data.email_verified` correctly handles both `false` and `undefined`.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Code refactor
- [ ] Documentation update

## Related issue

Closes #1527

N/A.

## Checklist


Anexus5919:
Thanks!

Anexus5919:
@ronisarkarexe Heads up: while this PR was open, the same `email_verified` guard for Google login appears to have already landed on `main` (same condition and same `UNAUTHORIZED` response). I have merged the latest `main` and resolved the conflict by keeping the single existing check rather than duplicating it, so the branch no longer adds a second copy.

Net effect: the `auth.service.ts` logic on this branch is now identical to `main`. The vulnerability is already addressed upstream, so this PR may be redundant and can be closed if you prefer. I am leaving it open for you to decide. Thanks for taking a look.

PR #811, opened 2026-05-28
fix(auth): token versioning + change password

Rishitgoyal06:
Fixes #635

## What this PR does

Adds per-user tokenVersion and includes it in access/refresh JWT payloads.

The auth middleware and refresh-token flow now reject tokens when the user’s current tokenVersion doesn’t match the token payload, invalidating previously issued tokens.

Adds a protected POST /auth/change-password endpoint that:
- verifies old password
- updates password
- increments tokenVersion
- invalidates all previous sessions

## Type of change
- Bug fix
- Security improvement

## Notes
I performed local code-level verification and reviewed the authentication flow, but full end-to-end testing was limited due to MongoDB connection setup issues.

## Checklist
- [x] Self-reviewed code
- [x] Kept changes scoped to auth/session logic
- [x] Maintained existing refresh-token flow

Issue #635, opened 2026-05-27
[BUG] : Session Invalidation on Password Reset (Backend Security Bug)


hrshjswniii:
# Problem
---

Currently, when a user resets or changes their password, the backend correctly:

* hashes and stores the new password
* issues fresh JWT tokens for auto-login

However, all previously issued JWT access/refresh tokens remain fully valid until their expiration time.

This creates a serious security vulnerability:

* if an attacker already possesses a valid token
* and the legitimate user changes their password after compromise

…the attacker can continue accessing the account using old JWTs even after the password reset succeeds.

At the moment:

* JWTs are stateless
* no token revocation mechanism exists
* auth middleware never checks whether tokens were invalidated after a security-sensitive event

This means password reset currently does **not actually terminate existing sessions**.

---
---

# Proposed Solution
---

Implement lightweight JWT session invalidation using a database-backed **Token Versioning** strategy.

Instead of introducing Redis-based blacklist infrastructure, this approach:

* remains simple
* scales well
* integrates cleanly into the existing auth architecture

---
---

# Solution Overview
---

### 4. Enforce Token Version Checks in Auth Middleware

---
---

# Why This Approach?
---

Compared to Redis/JWT blacklists:

| Token Versioning          | Redis Blacklist                  |
| ------------------------- | -------------------------------- |
| No new infrastructure     | Requires Redis                   |
| Minimal code changes      | Operational overhead             |
| Database-backed           | Cache synchronization complexity |
| Stateless JWTs preserved  | Stateful blacklist tracking      |
| Simple horizontal scaling | More moving parts                |

This makes it ideal for the current architecture.

---
---

# Files Affected
---

| File                                             | Changes                                                                    |
| ------------------------------------------------ | -------------------------------------------------------------------------- |
| `backend/src/app/modules/user/user.interface.ts` | Add `tokenVersion` to `IUser`                                              |
| `backend/src/app/modules/user/user.model.ts`     | Add `tokenVersion` schema field                                            |
| `backend/src/app/modules/auth/auth.service.ts`   | Include `tokenVersion` in JWT payloads and increment during password reset |
| `backend/src/app/middlewares/auth.middleware.ts` | Validate JWT `tokenVersion` against DB                                     |
| JWT generation utilities                         | Ensure all auth flows propagate version                                    |

---
---

# Acceptance Criteria
---

| Criteria                                                        | Status |
| --------------------------------------------------------------- | ------ |
[...]

PR #1537, opened 2026-05-30
feat(auth): add refresh token rotation with reuse detection and logout

### LINKED ISSUE #1536: BUG: No refresh-token rotation, and password reset does not invalidate old sessions

### What happened?

Two related auth weaknesses around refresh tokens.

1. No refresh-token rotation or reuse detection. `POST /api/v1/auth/refresh-token` (`auth.service.ts` `refreshToken`) only mints a new access token. The long-lived refresh token is never rotated, and there is no server-side record of issued refresh tokens, so a stolen refresh token can be replayed for its full lifetime and the theft cannot be detected. RFC 6819 section 5.2.2.3 recommends rotating the refresh token on each use and treating reuse of a retired token as a compromise signal.

2. Password reset does not invalidate existing sessions, and its auto-login tokens are broken. The token versioning mechanism added in PR #811 (issue #635) handles `change-password`, but the forgot-password `resetPassword` path was left out:
   - It never increments `tokenVersion`, so sessions that existed before the reset stay valid. This is exactly the vulnerability #635 set out to fix (step 3, "Invalidate Existing Sessions on Password Reset"), still present for the reset path.
   - The auto-login tokens it returns omit `tokenVersion` entirely. The auth middleware compares `user.tokenVersion` (0) against the token's `tokenVersion` (undefined), so `0 !== undefined` is true and the very next protected request returns 401. Auto-login after a password reset is currently broken.

This maps to CWE-613 (Insufficient Session Expiration) and CWE-384 (Session Fixation / reuse).

### Steps to reproduce

Rotation:
1. Log in. Call `POST /api/v1/auth/refresh-token`. Only a new access token is returned; the same refresh token keeps working indefinitely, with no rotation and no reuse detection.

Reset path:
1. Complete the forgot-password flow and reset the password.
2. The response includes an access token, but the first protected API call with it returns 401 because the token has no `tokenVersion`.
3. Separately, any token issued before the reset still works, so the reset did not end existing sessions.

### Environment (if relevant)

- Files: `backend/src/app/modules/auth/auth.service.ts`, `auth.controller.ts`, `auth.router.ts`, `backend/src/utils/cookie.util.ts`

### Suggested fix

- Persist issued refresh tokens with a unique id (jti) in a MongoDB collection with a TTL index.
- On refresh, rotate: validate the token, mark its record used, issue a new access and a new refresh token, and set the new refresh cookie. If a refresh token that was already used is presented again, treat it as theft and revoke the whole family plus bump `tokenVersion`.
- Read the refresh token from the httpOnly cookie (it is set as a cookie, but the endpoint was reading the Authorization header, which the client cannot supply for an httpOnly cookie).
- In `resetPassword`, increment `tokenVersion` and include it in the issued tokens, so reset both invalidates old sessions and produces working auto-login tokens.
- Add a `logout` endpoint that revokes the current refresh session and clears the cookie.

A PR implementing all of this is attached.

N/A. These are code-level auth weaknesses.

### Record

Anexus5919:
## Screenshot (when helpful)

N/A. Backend auth feature. Verified via type check and a code trace, described below.

## What this PR does

Adds stateful refresh-token rotation with reuse detection, fixes the password-reset session invalidation gap, and adds a logout endpoint.

### Background

Token versioning (PR #811, issue #635) already covers change-password and is enforced in the auth middleware. This PR builds on it to add true refresh-token rotation (RFC 6819 5.2.2.3) and to close the gaps that remained.

### Changes

- New `refresh_session.model.ts`: a `RefreshSession` collection (`jti`, `userId`, `used`, `revoked`, `expiresAt`) with a TTL index, so each issued refresh token is tracked and expired records are cleaned up automatically.
- `auth.service.ts`:
  - Added `issueAccessToken` and `issueRefreshToken` helpers. The refresh helper mints a token with a unique `jti` and records a session. `login`, `register`, `googleLogin`, and `resetPassword` now issue tokens through these helpers, so every session is tracked and every token carries `tokenVersion`.
  - Rewrote `refreshToken` to rotate: verify the token, check `tokenVersion`, look up the session by `jti`, and atomically claim it (`used: false -> true`) so concurrent calls cannot double-rotate. It then issues a new access and a new refresh token. If a token that was already used is presented again, that is treated as theft: the whole family is revoked and `tokenVersion` is bumped to kill access tokens too.
  - Added `logout`, which revokes the presented refresh session.
  - `resetPassword` now increments `tokenVersion` and revokes outstanding sessions, so a reset ends old sessions, and the auto-login tokens it returns now include `tokenVersion` (fixing a live bug where post-reset tokens 401'd immediately because the middleware compared `0 !== undefined`).
- `auth.controller.ts`: `refreshToken` now reads the refresh token from the httpOnly cookie (it is set as a cookie, but the endpoint was reading the Authorization header, which a client cannot supply for an httpOnly cookie) and sets the rotated cookie. Added the `logout` controller.
- `auth.router.ts`: added `POST /auth/logout`.
- `cookie.util.ts`: added `clearRefreshTokenCookie`.

### How I verified

1. Traced each flow: login/register/google/reset create a tracked session and tokens carrying `tokenVersion`; refresh rotates and detects reuse; logout revokes; reset bumps `tokenVersion` and revokes sessions.
2. Confirmed the atomic `findOneAndUpdate({ jti, used: false, revoked: false }, { used: true })` makes concurrent rotation safe and underpins reuse detection.
3. Ran `tsc` on the backend. All changed files compile cleanly. The only remaining errors are the pre-existing ones in `story_version.service.ts`, unrelated to this PR.

### Notes

- Mongo-backed (the project already requires MongoDB), so the session state is shared across serverless instances.
- The frontend currently has no refresh call wired (it stores the access token in localStorage and never calls `/refresh-token`). This PR makes the backend correct and ready; wiring the frontend to call `/refresh-token` (with credentials) on 401 and to call `/logout` is a separate frontend task.

## Type of change

- [x] Bug fix
- [x] New feature
- [ ] Code refactor
- [ ] Documentation update

## Related issue

Closes #1536

N/A.

## Checklist


Anexus5919:
@ronisarkarexe Kindly have a review on this pr whenever you get time.
Thanks!

Issue #1149, opened 2026-05-29
BUG: Story Continuation REST Endpoint Returns Hardcoded Placeholder (Unprotected)


nyxsky404:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### What happened?

The `/api/v1/story-continuation/continue` POST endpoint in `backend/src/routes/story.routes.ts` returns a hardcoded placeholder string instead of calling any AI service:

```ts
router.post("/continue", async (req, res) => {
  try {
    const { prompt } = req.body;

    // Replace this with existing AI generation logic
    const generatedText = "This is the generated continuation chapter."; // ← placeholder

    res.json({ text: generatedText });
  } catch (error) {
    res.status(500).json({ error: "Failed to continue story" });
  }
});
```

Additionally:
- The endpoint has **no authentication middleware** — anyone can call it
- There is **no input validation** on `req.body`
- The `prompt` field is destructured but never used
- The same route file is mounted in **two places**: `/review/continue` (in `app.ts`) and `/api/v1/story-continuation/continue` (in `router/index.ts`)
- The `/create` route in the same file is equally non-functional (logs to console and returns a hardcoded success)

**Expected:** The endpoint should either call the existing AI service (e.g., `AiModelService` / Gemini utilities) to generate a real continuation, or be removed until it is ready.

**Actual:** Returns the literal string `"This is the generated continuation chapter."` for every request regardless of input.

> **Note:** This is distinct from #771 (which covers the `collab:ai_continue` WebSocket handler in `collab.socket.ts`). This issue is about the REST endpoint in `story.routes.ts`.

### Steps to reproduce

1. Start the backend server
2. Send a POST request to `/api/v1/story-continuation/continue` with any body (or no body at all):
   ```bash
   curl -X POST http://localhost:<port>/api/v1/story-continuation/continue \
     -H "Content-Type: application/json" \
     -d '{"prompt": "anything"}'
   ```
3. Observe the response is always: `{"text": "This is the generated continuation chapter."}`
4. Note that no authentication token is required

### Environment (if relevant)

- File: `backend/src/routes/story.routes.ts`
- Route registration: `backend/src/router/index.ts` (line with `/story-continuation`) and `backend/src/app.ts` (line with `/review`)
- Affects: All environments (dev, staging, production)

### Screenshots or screen recording (if helpful)

N/A — reproducible via any HTTP client (curl, Postman, browser).

### Record

[...]

PR #1540, opened 2026-05-30
fix: remove duplicate /review mount referencing unimported storyRoutes

### LINKED ISSUE #1273: Duplicate route registration for story.routes
app.ts mounts storyRoutes at /review and router/index.ts also registers the same routes, creating duplicate endpoints. Worse, express.json() is added after app.use("/review", storyRoutes), so the body parser is not applied to those routes req.body will always be undefined for them. Fix by moving body-parser middleware before all routes.

Anexus5919:
## Screenshot (when helpful)

N/A. One-line backend fix. Verified by code trace and type check, described below.

## What this PR does

Removes the duplicate route mount in `app.ts` that also broke server startup.

`app.ts` had:

```
app.use("/review", storyRoutes);
```

Two problems with that line:

1. `storyRoutes` is not imported in `app.ts` (it is only imported in `router/index.ts`, where it is mounted at `/api/v1/story-continuation`). So this line references an undefined identifier. The dev server runs with `ts-node-dev --transpile-only`, which skips type checking, so this surfaces at runtime as `ReferenceError: storyRoutes is not defined` when `app.ts` loads, crashing the server on startup. (Plain `tsc --noEmit` does not flag it right now only because the pre-existing syntax errors in `story_version.service.ts` short-circuit full semantic analysis.)
2. It registered the same router a second time. `storyRoutes` is already mounted at `/api/v1/story-continuation` via `router/index.ts`, so `/review` was a duplicate registration. This is the issue reported in #1273.

Deleting the line fixes the startup crash, removes the duplicate registration, and drops the unreachable no-op `POST /review/create` (which logged the body and returned 201 without persisting anything).

The legitimate `/api/v1/story-continuation` mount is untouched. The hardcoded placeholder in that router (`/continue`) is a separate concern tracked in #1149 and is intentionally left for that issue.

## How I verified

1. Confirmed `storyRoutes` is imported only in `router/index.ts`, never in `app.ts`, and that no global declaration defines it, so the `app.ts` usage was an undefined reference.
2. After the change, `app.ts` references only imported identifiers, and `/api/v1/story-continuation` remains wired in `router/index.ts`.
3. Ran `tsc` on the backend. No new errors. The only remaining errors are the pre-existing ones in `story_version.service.ts`, unrelated to this PR.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Code refactor
- [ ] Documentation update

## Related issue

Closes #1273

N/A.

## Checklist


Anexus5919:
@ronisarkarexe Kindly have a review on this pr whenever you get time.
Thanks!

Anexus5919:
Heads up: while this PR was open, the duplicate `app.use("/review", storyRoutes)` line was already removed on `main`. I have merged the latest `main` and resolved the conflict, and the `app.ts` on this branch is now identical to `main`.

So the duplicate mount issue (#1273) is already addressed upstream and this PR is effectively a no-op now. It can be closed if you prefer; I am leaving it open for you to decide. Thanks!

PR #1556, opened 2026-05-31
Fix #1555: Remove hardcoded default credentials in login.html

### LINKED ISSUE #1555: Remove hardcoded default credentials in login.html
In `login.html`, the email and password fields have default values hardcoded into the HTML:

- Email field (around line 122): `value="admin@gmail.com"`
- Password field (around line 134): `value="admin@123"`

This is a security risk and bad UX for a production application, as real users will have to clear these credentials before entering their own. Please remove the `value` attributes from these inputs so the fields start empty.

assign me under gssoc 26

Goutam-2702:
Resolves #1555 by removing the hardcoded value attributes for email and password inputs in login.html.


ronisarkarexe:

PR #1564, opened 2026-05-31
fix(audio): restart utterance with new rate on speed change

### LINKED ISSUE #1327: BUG:  Playback Speed Control Does Not Work  Audio Plays at Normal Speed Regardless of Setting

### What happened?

The playback speed selector on the Stories page has no effect on audio playback. Even when a different speed (e.g., **2×**) is selected, the story audio continues to play at the default **1× (normal)** speed. The UI control updates visually but does not actually change the playback rate of the audio.
 

### Steps to reproduce

1. Go to https://storysparkai.vercel.app/stories
2. Open any story and start audio playback.
3. Locate the **playback speed** control (e.g., `1×`, `1.5×`, `2×` selector).
4. Select a speed other than `1×` - for example, **2×**.
5. Listen to the audio.

### Environment (if relevant)

_No response_

_No response_

### Record

shrutisharma-sh:
## What this PR does
The playback speed selector on the Stories page had no effect on audio
playback. Selecting 2× (or any speed) updated the UI visually but the
narration continued at 1× speed.

**Root cause:** The Web Speech API silently ignores `rate` changes on an
already-speaking `SpeechSynthesisUtterance`. Setting
`utterance.rate = newRate` after `speak()` has been called does nothing.

**Fix:** In `useSpeechSynthesis.ts`, the `setRate` function now:
1. Cancels the active utterance when speed is changed mid-playback
2. Slices the remaining text from the current word index
3. Creates a new utterance with the updated rate and resumes from that point
4. Tracks `currentWordIndex` in a ref so the callback always has a fresh
   value without a stale closure

If narration is paused when speed changes, playback stops gracefully since
resuming mid-utterance with a new rate is not reliable cross-browser.

**To verify the fix:**
1. Go to `/stories`, generate or open a story
2. Start audio playback
3. Change speed to 2× — narration should immediately restart at double speed
4. Change back to 1× — narration should restart at normal speed

## Type of change
- [x] Bug fix

## Related issue
Closes #1327 

## Checklist

### DIFF
--- a/frontend/src/hooks/useSpeechSynthesis.ts
+++ b/frontend/src/hooks/useSpeechSynthesis.ts
@@ -202,7 +202,9 @@ export const useSpeechSynthesis = (text: string): UseSpeechSynthesisResult => {
       }
 
       if (typeof event.charIndex === "number") {
-        setCurrentWordIndex(getWordIndexAtCharIndex(event.charIndex, wordRanges));
+        const idx = getWordIndexAtCharIndex(event.charIndex, wordRanges);
+        setCurrentWordIndex(idx);
+        currentWordIndexRef.current = idx;
       }
     };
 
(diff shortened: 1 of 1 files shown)


ronisarkarexe:

Issue #1191, opened 2026-05-29
Issue: NoSQL Injection Vulnerability in Story Creation/Updates


PiyushBorban49:
# Secure Payload Validation for Story/Post APIs

## Description

The current implementation of story/post creation and update operations directly passes `req.body` to the service layer and subsequently to the Mongoose models without strict runtime schema validation at the controller level.

While Mongoose provides some schema enforcement, passing unfiltered objects introduces several security risks:

### Potential Risks

* **Query Operator Injection**

  * Attackers may send payloads containing MongoDB/Mongoose operators such as:

    * `$ne`
    * `$gt`
    * `$where`
  * These operators can be injected into fields expected to contain plain strings or primitive values, potentially bypassing business logic or exposing unintended data.

* **Mass Assignment Vulnerability**

  * Blindly merging request payloads may allow modification of protected/internal fields such as:

    * `isFeaturedPost`
    * `author`
    * `isDeleted`
    * other restricted properties

---

### 1. Schema Validation with Zod

Define strict Zod schemas for:

* `createPost`
* `updatePost`

This ensures:

* Runtime payload validation
* Type-safe request handling
* Rejection of unexpected fields
* Improved API consistency

---

### 2. Middleware Integration

Apply the `validateRequest` middleware in `post.router.ts` for all relevant routes.

Example:

```ts
router.post(
  "/create-post",
  validateRequest(createPostSchema),
  PostController.createPost
);

router.patch(
  "/:id",
  validateRequest(updatePostSchema),
  PostController.updatePost
);
```

---

### 3. Strict Field Filtering

Only extract and forward explicitly allowed fields from `req.body` before interacting with the database layer.

Example:

```ts
const payload = {
  title: req.body.title,
  content: req.body.content,
  category: req.body.category,
};
```

Avoid directly passing:

```ts
Post.create(req.body);
```

---

## Expected Outcome

* Prevent MongoDB operator injection
* Mitigate mass assignment vulnerabilities
* Enforce strict API contracts
* Improve overall request validation and application security

awhsaY4444:
Hi @ronisarkarexe, I'd like to work on this issue under GSSoC'26. I have reviewed the requirements and would like to implement the Zod validation schemas, middleware integration, and strict payload filtering. Could you please assign this issue to me?

ARPANPATRA111:
I'd like to work on this — contributing under GSSoC'26

PR #1649, opened 2026-05-31
fix: import validateRequest and PostValidator and validate create-post body

Anexus5919:
## Screenshot (when helpful)

N/A. Small backend fix. Verified via type check and a code trace, described below.

## What this PR does

Repairs the post payload validation wiring so it actually runs.

`post.router.ts` calls `validateRequest(PostValidator.updatePost)` on the `PATCH /:id` route, but neither `validateRequest` nor `PostValidator` is imported in that file. At runtime that is a `ReferenceError` on the update-post route (the dev server uses `ts-node-dev --transpile-only`, so it is not caught at build time, and a separate pre-existing parse error in `review.service.ts` currently masks it from `tsc`). The create-post route also has no validation at all.

This change:
- Imports `validateRequest` (from `../../middleware/validate.request`) and `PostValidator` (from `./post.validation`), so the existing `updatePost` validation works instead of throwing.
- Adds `validateRequest(PostValidator.createPost)` to the `POST /create-post` route so create payloads are validated too.

Together this makes the post create and update endpoints reject malformed or operator-laden bodies (for example a `content` value of `{ "$ne": null }`) before they reach Mongoose, which is the validation hardening described in #1191.

## How I verified

1. Confirmed `validateRequest` and `PostValidator` were used on the PATCH route but never imported.
2. Confirmed both symbols exist with the expected paths and shapes (the auth router imports them the same way).
3. Ran `tsc` on the backend. With the unrelated pre-existing parse error in `review.service.ts` temporarily set aside, `post.router.ts` type-checks cleanly with the imports added, and no new errors are introduced. The only remaining error on the branch is that pre-existing `review.service.ts` one, which is unrelated to this change.

## Related issue

Relates to #1191 (post create/update payload validation).

## Note for the maintainer

While verifying this I noticed two separate pre-existing issues that this PR intentionally does not touch: a parse error in `backend/src/app/modules/review/review.service.ts` (line 83) and a leftover `Cannot find name 'main'` merge artifact in `backend/src/app/modules/comment/comment.service.ts` (line 46). Both break the backend build independently and are worth their own fixes.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Code refactor
- [ ] Documentation update

N/A.

## Checklist

### DIFF
--- a/backend/src/app/modules/post/post.router.ts
+++ b/backend/src/app/modules/post/post.router.ts
@@ -2,6 +2,8 @@ import express from "express";
 import { PostController } from "./post.controller";
 import auth from "../../middleware/auth.middleware";
 import checkRequestLimit from "../../middleware/check.request.limit";
+import validateRequest from "../../middleware/validate.request";
+import { PostValidator } from "./post.validation";
 import { ENUM_USER_ROLE } from "../../../enums/user";
 
 const router = express.Router();
@@ -13,6 +15,7 @@ const router = express.Router();
 router.post(
   "/create-post",
   auth(),
+  validateRequest(PostValidator.createPost),
   PostController.createPost
 );
 


Anexus5919:
@ronisarkarexe kindly have a review on this pr. Thanks!

Anexus5919:
@ronisarkarexe Could you please consider increasing the GitHub assignment limit or removing it altogether? It becomes difficult to manage when issues get assigned without considering the time and effort already invested in auditing, researching, and working on them. A more flexible assignment policy would help avoid duplicate work and improve coordination among contributors.

PR #1654, opened 2026-05-31
fix: restore comment retrieval service

### LINKED ISSUE #1652: BUG: backend build fails because comment service contains stray main token
Line: `backend/src/app/modules/comment/comment.service.ts:46`

I ran into this while checking the latest backend build state after rebasing another branch. The backend currently has a stray `main` token inside `getCommentsByPostId`, so TypeScript cannot compile the service.

Problem:
`getCommentsByPostId` is effectively empty and contains only:

```ts
const getCommentsByPostId = async (postId: string) => {
main
};
```

That makes `tsc` fail with:

```text
src/app/modules/comment/comment.service.ts(46,1): error TS2304: Cannot find name 'main'.
```

Impact:
This blocks the backend build on the current branch. It also means the comment listing path cannot be trusted because the service method implementation has been replaced by an invalid token.

How to verify:

1. Check `backend/src/app/modules/comment/comment.service.ts` around line 46.
2. Run the backend build or wait for CI.
3. TypeScript stops at the stray `main` token before the backend can compile.

Suggested fix:
Restore the intended `getCommentsByPostId` implementation instead of the stray token. It should fetch comments for the post, probably using the existing `Comment` model, returning the expected DTO/lean shape used by the controller.

Labels: `type:bug`, `gssoc'26`, likely `level:intermediate`

Raising this as part of GSSoC 2026. I searched existing issues for the same `comment.service.ts` / `Cannot find name 'main'` build failure and did not find a duplicate.

eshaanag:
Screenshot (when helpful)
N/A - backend service build fix only; no UI change.

What this PR does
This fixes the backend compile break reported in #1652.

`getCommentsByPostId` in `comment.service.ts` currently contains a stray `main` token, which makes TypeScript fail with `Cannot find name 'main'`. This PR restores the comment retrieval implementation so the service:

- validates that the post exists and is not deleted
- fetches comments for the post with user and likes populated
- groups replies under their parent comments
- returns the existing `{ comments, totalComments }` response shape

Type of change
- [x] Bug fix
- [ ] New feature
- [ ] Code refactor
- [ ] Documentation update

Related issue
Closes #1652

More screenshots (optional)
N/A

Checklist

Local checks
- `git diff --check` passed.
- `npm run build --workspace backend` could not complete locally because `tsc` is not installed in this checkout.

### DIFF
--- a/backend/src/app/modules/comment/comment.service.ts
+++ b/backend/src/app/modules/comment/comment.service.ts
@@ -43,39 +43,55 @@ const createComment = async (
 };
 
 const getCommentsByPostId = async (postId: string) => {
+  const post = await Post.findOne({ _id: postId, isDeleted: { $ne: true } });
+  if (!post) {
+    throw new ApiError(httpStatus.NOT_FOUND, "Post not found!");
+  }
+
   const allComments = (await Comment.find({ postId })
     .populate("userId", "name email")
     .populate("likes")
     .sort({ createdAt: -1 })
-    .lean()) as any[];
+    .lean()) as unknown as ILeanComment[];
 
   const totalComments = allComments.length;
 
-  const topLevelComments: any[] = [];
-  const replyMap = new Map<string, any[]>();
+  const topLevelComments: ICommentDTO[] = [];
+  const replyMap = new Map<string, ICommentDTO[]>();
 
-  // Distribute comments into top-level list and replies map
   for (const comment of allComments) {
-    if (!comment.parentCommentId) {
-      comment.replies = [];
-      topLevelComments.push(comment);
+    const commentDTO: ICommentDTO = {
+      ...comment,
+      replies: [],
+    };
+
+    if (!commentDTO.parentCommentId) {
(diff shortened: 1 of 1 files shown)


eshaanag:
Rechecked this after rebasing the branch on the latest `main`.

Current status:

- `lint` is passing
- `typecheck` is passing
- the PR is mergeable with no conflicts
- the only remaining failing check is `CI / build`

The build job is failing before project dependencies install or any app code runs:

```text
Unable to locate executable file: pnpm
```

That comes from `.github/workflows/main.yml`, where `actions/setup-node` is configured with `cache: pnpm`, but the workflow does not install pnpm first and then runs `npm install` afterward. So the current failure is in the CI setup, not in `backend/src/app/modules/comment/comment.service.ts`.

I kept this PR scoped to the comment service fix for #1652. If maintainers want, I can handle the workflow setup separately, but I did not want to mix a CI workflow change into this bugfix PR without review.

Issue #1534, opened 2026-05-30
fix(ci): pnpm-lock.yaml is out of sync with package.json causing CI installation failures


Krish00711:
## Description

The CI workflow currently fails during dependency installation because `pnpm-lock.yaml` appears to be out of sync with `package.json`.

### Observed Behavior

The CI logs report:

```text
ERR_PNPM_OUTDATED_LOCKFILE

Cannot install with "frozen-lockfile" because pnpm-lock.yaml is not up to date with package.json
```

The workflow identifies dependency specifier mismatches between the lockfile and package manifests.

### Impact

* CI checks fail even when unrelated code changes are submitted.
* Contributors may receive failing checks unrelated to their modifications.
* Linting and validation workflows cannot complete successfully.

### Expected Behavior

* `pnpm-lock.yaml` should accurately reflect the current dependency definitions in `package.json`.
* CI workflows using `--frozen-lockfile` should complete successfully.

### Suggested Fix

* Regenerate and commit an updated `pnpm-lock.yaml`.
* Verify lockfile consistency across all workspaces.
* Ensure CI dependency installation succeeds with `--frozen-lockfile`.

### Benefits

* More reliable CI pipelines.
* Improved contributor experience.
* Reduced false-negative build failures.

Krish00711:
Hi maintainers 👋

I am currently working on a higher-priority build restoration issue that affects repository compilation.

To avoid holding multiple assignments at once, I am releasing this issue for now.

Thank you!

/unassign

karuna-1:
Hi, I was trying to set up the project locally by following the README instructions.

When I ran `npm install`, I got the following error:

npm error code EOVERRIDE
Override for @types/express@^5.0.6 conflicts with direct dependency

My environment:

* macOS
* Node.js v24.15.0
* npm v11.12.1

I also noticed the repo has `pnpm-lock.yaml`, `pnpm-workspace.yaml`, and `pnpm@9.15.9` specified in `package.json`, but the README mentions using `npm install`.

Could this be connected to the issue being discussed here, or am I missing a setup step?

PR #1665, opened 2026-05-31
fix: remove dead monthly-reset cron that never runs on serverless

### LINKED ISSUE #1664: BUG: Monthly quota-reset cron is dead code on Vercel and is redundant with the lazy reset

### What happened?

`backend/src/app.ts` schedules a monthly cron to reset `requestsThisMonth`, but it is wrapped in `if (!process.env.VERCEL)`, so it never runs on the Vercel serverless deployment. Even off Vercel it would not run reliably, because serverless and most process-per-request hosts do not keep a long-lived process alive to fire `node-cron`.

```ts
if (!process.env.VERCEL) {
  cron.schedule("0 0 1 * *", async () => {
    await User.updateMany({}, { $set: { requestsThisMonth: 0 } });
  });
}
```

The important part: this cron is not actually needed. The real monthly reset is already handled lazily and correctly inside `reserveUserQuota` (`quota.service.ts`). Its `effectiveCountExpression` uses a `$cond` that treats `requestsThisMonth` as 0 whenever `lastRequestDate` is null or older than the first day of the current month. So a user whose stored count is left over from last month is correctly treated as having 0 used this month the moment they next generate, regardless of the cron. `requestsThisMonth` is never read directly for enforcement or shown to users (it is only referenced in tests), so the stored value being stale has no effect.

The legacy non-atomic reset that used to live in `ai_model.service.ts` has already been removed, so the cron is the only remaining reference to the old approach and is now misleading dead code: it implies a scheduled reset happens when, on the actual deployment, it does not.

### Steps to reproduce

1. Deploy on Vercel (or any serverless host). `process.env.VERCEL` is set, so the cron is never scheduled.
2. Confirm via `reserveUserQuota` that the monthly limit still resets correctly anyway, because `effectiveCountExpression` zeroes last-month usage on the next request.

### Environment (if relevant)

- File: `backend/src/app.ts`
- Quota logic: `backend/src/app/modules/ai_model/quota.service.ts`

### Suggested fix

Remove the dead cron block and its now-unused imports (`node-cron`, `User`, `NewsletterSubscriber`) from `app.ts`. The lazy `$cond` reset in `reserveUserQuota` is authoritative and unchanged, so enforcement behavior is identical. A separate Vercel Cron Job would only be worth adding if a stored, eagerly-reset counter is ever needed for display, which is not the case today.

A PR removing the dead cron is attached. The `node-cron` dependency removal is left out of this PR to avoid a lockfile change while the lockfile is already out of sync (tracked in #1161 / #1534); it can be dropped from `package.json` in that cleanup.

N/A.

### Record

Anexus5919:
## Screenshot (when helpful)

N/A. Dead-code removal. Verified via type check and a code trace, described below.

## What this PR does

Removes the monthly quota-reset cron from `app.ts`, which never runs on the deployed environment and is redundant with the existing lazy reset.

The cron was guarded by `if (!process.env.VERCEL)`, so it never schedules on the Vercel serverless deployment (and would not fire reliably on any process-per-request host). More importantly it is not needed: `reserveUserQuota` in `quota.service.ts` already resets the monthly count lazily and correctly. Its `effectiveCountExpression` uses a `$cond` that treats `requestsThisMonth` as 0 whenever `lastRequestDate` is null or before the first day of the current month, so the limit resets on the user's next request regardless of any cron. `requestsThisMonth` is never read for enforcement or shown to users (only referenced in tests), so a stale stored value is harmless.

This removes the cron block and its now-unused imports (`node-cron`, `User`, `NewsletterSubscriber`). Enforcement behavior is unchanged.

## How I verified

1. Confirmed `reserveUserQuota` is the only quota gate and that its `$cond` handles the month rollover; the old non-atomic reset in `ai_model.service.ts` has already been removed.
2. Confirmed `cron`, `User`, and `NewsletterSubscriber` were used only by the removed block (`NewsletterSubscriber` was already unused), and that `node-cron` is not imported anywhere else.
3. Ran `tsc` on the backend. With the unrelated pre-existing errors in `comment.service.ts` and `recommendation.service.ts` set aside, `app.ts` type-checks cleanly. I confirmed those two errors are present on clean `main` as well, so they are not introduced here.

## Note

I left `node-cron` in `package.json` rather than removing it in this PR, to avoid a lockfile change while the lockfile is already out of sync (tracked in #1161 / #1534). It can be dropped during that cleanup. Separately, `comment.service.ts` (line 46, a `Cannot find name 'main'` artifact) and `recommendation.service.ts` (line 58) have pre-existing build errors worth their own fixes.

## Type of change

- [x] Bug fix
- [ ] New feature
- [x] Code refactor
- [ ] Documentation update

## Related issue

Closes #1664

N/A.

## Checklist


Anexus5919:
Thanks!

Anexus5919:
@ronisarkarexe Resolved the conflict by merging the latest `main`. Kindly have a review on it. Thanks!

PR #1674, opened 2026-05-31
refactor(bookmark): make Bookmark collection the single source of truth

### LINKED ISSUE #1673: BUG: Bookmarks have two sources of truth (Bookmark collection + Post.bookmarks array) that can drift

### What happened?

Bookmarks are stored in two places that must be kept in sync by hand:

- The `Bookmark` collection (`userId`, `storyId`), used by `bookmark.service.ts`.
- A `bookmarks` array of user ids on each `Post`, used by `post.service.ts` `toggleBookmark` and read by the frontend.

`bookmark.service.toggleBookmark` writes both: it creates/deletes a `Bookmark` document and then separately updates `Post.bookmarks`. Each operation is now individually atomic (an earlier fix replaced the old non-atomic `save()`), but the two writes are still not atomic together. If the process fails between them, the two stores diverge. There is also a second, parallel toggle (`post.service.toggleBookmark`, mounted at `POST /post/bookmark/:id`) that only updates `Post.bookmarks` and never touches the `Bookmark` collection, so the two endpoints can produce inconsistent state for the same post.

This is related to the already-resolved concurrency fixes in #1009 / #991, but those addressed atomicity of the individual counters, not the dual source of truth itself.

This maps to CWE-doubled-state / data-integrity (denormalized data kept consistent only by application code).

### Steps to reproduce

1. Bookmark a story through the bookmark endpoint, then inspect both `Bookmark` documents and that post's `bookmarks` array; they are kept in sync only by sequential writes.
2. Toggle the same post through `POST /post/bookmark/:id`, which updates only `Post.bookmarks`, and observe the `Bookmark` collection is now out of step.

### Environment (if relevant)

- Files: `bookmark.service.ts`, `post.service.ts`, `post.model.ts`, `user.service.ts`, plus the frontend `BookmarkButton` and bookmark-count UI.

### Suggested fix

Make the `Bookmark` collection the single source of truth and replace the `Post.bookmarks` array with a denormalized `bookmarksCount` integer:

- Both backend toggles operate on the `Bookmark` collection and atomically `$inc` `Post.bookmarksCount`.
- The frontend derives the per-user bookmarked state from the existing `GET /bookmarks/status/:storyId` endpoint instead of scanning an array, and reads `bookmarksCount` for display.
- A one-time migration backfills existing `Post.bookmarks` arrays into `Bookmark` documents and sets `bookmarksCount`.

A PR implementing this, including the migration script, is attached.

N/A.

### Record

Anexus5919:
## Screenshot (when helpful)

N/A. Backend + frontend data-model refactor. Verified via type checks and a code trace, described below.

## What this PR does

Removes the dual source of truth for bookmarks. Previously bookmarks lived both in the `Bookmark` collection and in a `Post.bookmarks` array of user ids, kept in sync by hand across two endpoints. This PR makes the `Bookmark` collection authoritative and replaces the array with a denormalized `bookmarksCount` integer.

### Backend
- `post.model.ts` / `post.interface.ts`: drop the `bookmarks` array, add `bookmarksCount: number` (default 0).
- `bookmark.service.ts`: `toggleBookmark` and `deleteBookmark` operate on the `Bookmark` collection and atomically `$inc` `Post.bookmarksCount` (decrement guarded by `bookmarksCount > 0`). The duplicate-key race is still handled as a no-op success. `getBookmarks` no longer populates the removed array.
- `post.service.ts`: the parallel `toggleBookmark` (mounted at `POST /post/bookmark/:id`) now uses the `Bookmark` collection too, so both endpoints share one source of truth. Removed the six `.populate("bookmarks", ...)` calls.
- `user.service.ts`: account deletion no longer `$pull`s from the removed array; `Bookmark.deleteMany({ userId })` already removes the user's bookmarks.

### Frontend
- `BookmarkButton.tsx`: derives the bookmarked state from the existing `GET /bookmarks/status/:storyId` endpoint via `useCheckBookmarkStatusQuery` (skipped for guests), instead of scanning a `bookmarks` array prop. The `bookmarks` prop is removed.
- `feature`, `post.details`, `post.view.list`: drop the now-removed `bookmarks={...}` prop.
- `community_spotlight.component.tsx`: bookmark count now reads `post.bookmarksCount`.
- `models/post.ts`: remove the `Bookmark` array type, add `bookmarksCount`.

### Migration
- `backend/scripts/migrate-bookmarks.ts`: idempotent one-time script that backfills existing `Post.bookmarks` arrays into `Bookmark` documents (upsert) and sets `bookmarksCount`, then unsets the legacy array. Mirrors the connection pattern of `seed-admin.ts`. Run once after deploy: `npx ts-node scripts/migrate-bookmarks.ts`.

## How I verified

1. Confirmed every reader of `Post.bookmarks` (two backend toggles, user cleanup, frontend button, community-spotlight count) is migrated; no `Post.bookmarks` array reference remains in backend or frontend.
2. Backend `tsc`: my changed files compile cleanly; only the two unrelated pre-existing errors remain (`comment.service.ts`, `recommendation.service.ts`).
3. Frontend `tsc`: the error count is identical before and after this change (the frontend has many pre-existing JSX parse errors on `main`), and `BookmarkButton.tsx` and `models/post.ts`, the files carrying the real logic change, are clean in both runs, so this PR introduces no new type errors.
4. The migration script is under `scripts/` which is outside the `tsc` build (same as `seed-admin.ts`), so it does not affect CI typecheck.

## Note for the maintainer

The frontend on `main` currently has many pre-existing JSX build errors unrelated to this PR (for example in `feature.component.tsx`, `community_spotlight.component.tsx`, `post.view.list.component.tsx`). They predate this change and are out of scope here. Please run `scripts/migrate-bookmarks.ts` once after merge to backfill existing data.

## Type of change

- [x] Bug fix
- [ ] New feature
- [x] Code refactor
- [ ] Documentation update

## Related issue

Closes #1673

N/A.

## Checklist

### DIFF
--- a/backend/scripts/migrate-bookmarks.ts
+++ b/backend/scripts/migrate-bookmarks.ts
@@ -1,0 +1,51 @@
+import mongoose from "mongoose";
+import { Post } from "../src/app/modules/post/post.model";
+import { Bookmark } from "../src/app/modules/bookmark/bookmark.model";
+import config from "../src/config";
+
+// Idempotent migration: backfill legacy Post.bookmarks into Bookmark docs + bookmarksCount.
+const migrateBookmarks = async () => {
+  await mongoose.connect(config.database_url as string);
+
+  // Read the legacy array directly; the Post schema no longer defines it.
+  const docs = await Post.collection
+    .find({ bookmarks: { $exists: true, $type: "array" } })
+    .project({ bookmarks: 1 })
+    .toArray();
+
+  let createdBookmarks = 0;
+  let updatedPosts = 0;
+
+  for (const doc of docs) {
+    const userIds: mongoose.Types.ObjectId[] = Array.isArray(doc.bookmarks)
+      ? doc.bookmarks
+      : [];
+
+    for (const userId of userIds) {
+      const res = await Bookmark.updateOne(
+        { userId, storyId: doc._id },
+        { $setOnInsert: { userId, storyId: doc._id } },
+        { upsert: true }
+      );
+      if (res.upsertedCount) createdBookmarks += 1;
+    }
+
+    const count = await Bookmark.countDocuments({ storyId: doc._id });
+    await Post.collection.updateOne(
+      { _id: doc._id },
+      { $set: { bookmarksCount: count }, $unset: { bookmarks: "" } }
+    );
+    updatedPosts += 1;
+  }
+
+  console.log(
+    `Migration complete. Posts updated: ${updatedPosts}, bookmarks created: ${createdBookmarks}.`
+  );
+  await mongoose.disconnect();
+};
+
+migrateBookmarks().catch(async (error) => {
+  console.error("Bookmark migration failed:", error);
+  await mongoose.disconnect();
+  process.exit(1);
+});
--- a/backend/src/app/modules/bookmark/bookmark.service.ts
+++ b/backend/src/app/modules/bookmark/bookmark.service.ts
@@ -20,42 +20,30 @@ const toggleBookmark = async (storyId: string, token: ITokenPayload) => {
     throw new ApiError(httpStatus.BAD_REQUEST, "Story not found!");
   }
 
-  // Atomically find and delete the bookmark if it exists
+  // Bookmark collection is the single source of truth; bookmarksCount mirrors it.
   const deletedBookmark = await Bookmark.findOneAndDelete({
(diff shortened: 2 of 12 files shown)


Anexus5919:
Thanks!

Anexus5919:
Thanks!

PR #1722, opened 2026-05-31
fix: make DNS resolver override opt-in via DNS_SERVERS env var

### LINKED ISSUE #1721: [BUG]: server.ts overrides global DNS resolvers unconditionally at import time
**Describe the bug**

`backend/src/server.ts` calls `dns.setServers(["1.1.1.1", "8.8.8.8"])` unconditionally at module import time. This is a process-wide global side effect that overrides the platform DNS resolver for every lookup the backend makes.

On Vercel and most hosts the platform resolver works fine, and forcing public resolvers can break resolution of private or internal hostnames if internal services are added later. The call was introduced in a general refactor commit, not as a deliberate DNS fix, and the `dns` import is used only for this one line. This is a configuration anti-pattern (unconditional global side effect at import), not a security hole.

**To Reproduce**

1. Open `backend/src/server.ts`.
2. The first executable statement after the imports is `dns.setServers(["1.1.1.1", "8.8.8.8"])`.
3. It runs on every boot regardless of environment, replacing the host resolver globally.

**Expected behavior**

The backend should use the platform DNS resolver by default. A custom resolver override should be opt-in (configurable), not forced on every deployment.

**Screenshots**

N/A.

**Desktop (please complete the following information):**
  - OS: N/A (server-side configuration)
  - Browser: N/A
  - Version: N/A

**Additional context**

Suggested fix: make the override opt-in via a new optional `DNS_SERVERS` env var (comma-separated). Only call `dns.setServers` when it is set; otherwise use the platform resolver. This removes the unconditional global side effect while keeping an escape hatch for environments that genuinely need custom DNS (for example certain MongoDB Atlas SRV setups). Document `DNS_SERVERS` as optional in `.env.example`. I have a PR ready for this.

Anexus5919:
## Description

Removes the unconditional global DNS override in `server.ts`. Previously `dns.setServers(["1.1.1.1", "8.8.8.8"])` ran at import time on every boot, a process-wide side effect that overrides the platform resolver for all lookups and could break private or internal hostname resolution.

It is now opt-in via a new optional `DNS_SERVERS` env var:

- `config/index.ts`: parse `DNS_SERVERS` (comma-separated) using the same list helper as `CORS_ORIGINS` (renamed `parseCorsOrigins` to the generic `parseList`).
- `server.ts`: call `dns.setServers(config.dns_servers)` only when `DNS_SERVERS` is set; otherwise use the platform resolver.
- `.env.example`: document `DNS_SERVERS` as optional.

Anyone who needs custom resolvers (for example certain MongoDB Atlas SRV setups) can set `DNS_SERVERS=1.1.1.1,8.8.8.8` to restore the old behavior; the default no longer forces it.

Verified: `dns` was imported only for this single call, and `tsc` on the backend shows `config/index.ts` and `server.ts` clean (only unrelated pre-existing errors remain).

## Related Issue

Closes #1721

## Type of Change

- [x] Bug fix
- [ ] New feature
- [x] Documentation update
- [ ] Code refactor
- [ ] Other (please describe):

## Checklist

- [x] I have tested my changes
- [x] I have updated documentation if necessary
- [x] My code follows the project's coding style
- [x] This PR does not introduce any breaking changes
- [x] I have linked the related issue above

### DIFF
--- a/backend/.env.example
+++ b/backend/.env.example
@@ -3,6 +3,8 @@ NODE_ENV=development
 PORT=5000
 # Optional: comma-separated frontend origins for CORS (e.g. http://localhost:4001)
 CORS_ORIGINS=http://localhost:4001
+# Optional: comma-separated custom DNS resolvers (e.g. 1.1.1.1,8.8.8.8); unset uses the platform resolver
+# DNS_SERVERS=
 
 # Database
 # MongoDB connection string (local or Atlas)
--- a/backend/src/config/index.ts
+++ b/backend/src/config/index.ts
@@ -3,9 +3,7 @@ import path from "path";
 
 dotenv.config({ path: path.join(__dirname, "../../.env") });
 
-const parseCorsOrigins = (
-  raw: string | undefined
-): string[] | undefined => {
+const parseList = (raw: string | undefined): string[] | undefined => {
   if (!raw?.trim()) return undefined;
   return raw
     .split(",")
@@ -26,7 +24,8 @@ export default {
     }
     return url;
   })(),
-  cors_origins: parseCorsOrigins(process.env.CORS_ORIGINS),
+  cors_origins: parseList(process.env.CORS_ORIGINS),
+  dns_servers: parseList(process.env.DNS_SERVERS),
   bcrypt_salt_rounds: (() => {
     const raw = process.env.SALT_ROUNDS;
     const parsed = raw ? Number(raw) : NaN;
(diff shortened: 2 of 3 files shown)


Anexus5919:

namithar18:
Hi @ @ronisarkarexe,

I'd like to work on this issue as part of GSSoC. I noticed there is an open PR, but some checks are failing. If needed, I can work on an alternative fix and submit a PR.

Thanks!

Anexus5919:

Anexus5919:

PR #1741, opened 2026-06-01
fix(backend): bound rate-limit store memory by pruning stale entries

### LINKED ISSUE #1740: BUG: rateLimitMiddleware `store` Map grows unbounded, leaking memory on long-running servers

## What happened?

`backend/src/middlewares/rateLimitMiddleware.ts` keeps an in-memory sliding-window `store: Map<string, WindowEntry>`, but the cleanup logic only filters **timestamps inside** each `WindowEntry` — it never removes the `WindowEntry` itself from the Map.

### Concrete reproduction

```ts
// backend/src/middlewares/rateLimitMiddleware.ts
const store = new Map<string, WindowEntry>();   // line 19
// ...
const entry = store.get(key) ?? { timestamps: [] };
entry.timestamps = entry.timestamps.filter((t) => now - t < WINDOW_MS);  // line 44 — only filters inner array
// ...
entry.timestamps.push(now);
store.set(key, entry);
```

Sequence:
1. 10,000 unique clients (different `req.ip` or `x-forwarded-for` values) each make 1 request.
2. After 60s (WINDOW_MS), every `entry.timestamps` array becomes empty after the filter.
3. **`store.size` is still 10,000** — empty entries are never deleted.
4. Repeat over days/weeks: the Map grows monotonically.

### Why this matters

- **Memory exhaustion**: each entry is small (~80–120 bytes), but at scale (1M unique IPs over a week) that's 80–120 MB of leak.
- **CPU regression**: `Map.get` is O(1) but iteration during garbage collection scans all entries. The longer the server runs, the more work each GC cycle does.
- **Hard to detect in dev**: only shows up after days/weeks of uptime.
- **The middleware's own API acknowledges this**: the `clearRateLimitStore()` helper exists specifically because the author knew the store needs explicit cleanup — but no one calls it in production.

## Steps to reproduce

1. Start the backend in production mode.
2. Run a small load generator: e.g. `for i in {1..10000}; do curl -H "X-Forwarded-For: 10.$((i/65536)).$((i/256)).$((i%256))" -X POST <AI-generation-route>; done` (route protected by `storyRateLimiter`).
3. Wait 60+ seconds.
4. In a Node REPL attached to the same process, `require('./dist/middlewares/rateLimitMiddleware').getRateLimitStatus('10.0.0.1')` — returns `count: 0` for ALL 10,000 keys, but `store.size` is still 10,000.
5. Compare process memory: it grew ~1 MB even though every entry is "logically empty".

## Expected behavior

Empty `WindowEntry` records should be removed so the Map doesn't grow unbounded. Two reasonable fixes:

1. **Inline deletion**: after filtering timestamps, if `entry.timestamps.length === 0`, `store.delete(key)`.
2. **Periodic sweep**: a `setInterval` that prunes empty entries every N minutes.
3. **TTL-based cache**: replace the Map with `lru-cache` (already a common dep in Node projects) with a 60s TTL — eliminates the leak structurally.

A defensive inline check (option 1) is the smallest, most surgical change.

## Environment (if relevant)

- Node.js 20.x
- `storyRateLimiter` middleware mounted on AI generation routes
- Long-running production server (>= 24h uptime)

N/A — reproducible purely from process inspection.

## Work I'm planning to do

I'm contributing under **GSSoC'26**. Plan:

1. Implement fix option 1 (inline deletion of empty entries) — smallest, most targeted change.
2. Add a small unit test that asserts `store.size` returns to baseline after all timestamps evict.
3. Verify with the load generator above.
4. PR against `main`, target ~10–25 lines, expect `level:beginner` per the repo's auto-label rules.

## Record

vipul674:
## Before you open this PR

- **Issue:** Closes #1740
- **Description:** Memory-leak fix in `rateLimitMiddleware` — see below.
- **Screenshots:** N/A (backend-only change; test output pasted below)

## Screenshot (when helpful)

N/A — no UI change. Test run output:

```
PASS src/__tests__/rateLimitMiddleware.test.ts
  storyRateLimiter — basic behaviour
    ✓ allows exactly 10 requests from the same IP
    ✓ blocks the 11th request with 429
    ✓ uses user.id as key for authenticated requests
    ✓ reports correct remaining count via getRateLimitStatus
  storyRateLimiter — X-Forwarded-For spoofing
    ✓ ignores X-Forwarded-For and keys on req.ip
    ✓ does not create separate buckets per X-Forwarded-For value
  storyRateLimiter — store management
    ✓ clearRateLimitStore(key) removes only that key
    ✓ prunes entries whose timestamps are all outside the window (memory-leak fix)

Test Suites: 1 passed, 1 total
Tests:       8 passed, 8 total
```

## What this PR does

**Problem.** `rateLimitMiddleware.ts` keeps a `Map<string, WindowEntry>` (`store`) where each entry is the sliding-window timestamp list for one rate-limit key (per-user `user.id` or per-IP `req.ip`). For every request, the middleware writes the entry back via `store.set(key, entry)` — but the code never deletes entries whose timestamps have *all* expired. On a long-running server, every unique key that ever hit the limiter stays in the map forever, even after the window has rolled over. The map grows unbounded and is never reclaimed.

**Fix.** Add `pruneStaleEntries(now)` that:

1. Throttles itself to **at most once per minute** (using a module-level `lastPruneAt` watermark) so it has zero steady-state cost on the hot path.
2. Iterates `store.entries()`, filters timestamps to the active window, and:
   - **deletes** entries with no active timestamps (the actual leak),
   - **rewrites** entries where some timestamps are stale (keeps the map tidy, prevents the filter+push from running over long stale arrays).

The function is called at the very top of `storyRateLimiter`, *before* the per-request window filter, so it piggy-backs on real traffic — no background timer, no test-isolation headaches, no impact on hot-path latency beyond a single integer compare.

A debug log (`rateLimitMiddleware: pruned N empty entries, trimmed M entries`) is emitted via the existing `logger` when work was done, so it stays silent on idle servers.

**Test.** New unit test `prunes entries whose timestamps are all outside the window (memory-leak fix)` anchors `Date.now()` to a real timestamp, advances by 120 s, and asserts that the stale entry is **deleted** (not just stale-but-present), distinguishing the two states via `getRateLimitStatus(...).resetInMs` (0 if deleted, `WINDOW_MS` if stale-but-present).

**Files touched.**

- `backend/src/middlewares/rateLimitMiddleware.ts` — +46 / −4 lines
- `backend/src/__tests__/rateLimitMiddleware.test.ts` — +31 / 0 lines

## Type of change

- [x] Bug fix
- [x] Code refactor
- [ ] New feature
- [ ] Documentation update

## Related issue

Closes #1740

N/A

## Checklist

- [x] I have tested my changes (`npx jest src/__tests__/rateLimitMiddleware.test.ts` — 8/8 pass; full backend suite has 5 pre-existing failures in `ai_model.service.test.ts` and `quota.integration.test.ts` that are unrelated to this change and reproduce on `main`).

---

### Contributor note

This PR is submitted as part of **GirlScript Summer of Code 2026 (GSSoC'26)**. I would appreciate the **`gssoc'26`** label and a difficulty rating — this is a small, well-scoped backend bug fix that does not require knowledge of the broader codebase, so I would suggest **`level:beginner`** (PR < 100 lines, single concern, no new public API).

AI tool disclosure: I used an LLM as a research/code-review assistant to help identify the memory leak in the existing sliding-window implementation and to draft this PR description. All code, test, and review decisions were made by me.

### DIFF
--- a/backend/src/__tests__/rateLimitMiddleware.test.ts
+++ b/backend/src/__tests__/rateLimitMiddleware.test.ts
@@ -121,4 +121,35 @@ describe("storyRateLimiter — store management", () => {
     storyRateLimiter(r2, s2, n2);
     expect((s2 as any).statusCode).toBe(429);
   });
+
+  it("prunes entries whose timestamps are all outside the window (memory-leak fix)", () => {
+    // Bug: store grew unbounded because entries with only stale timestamps
+    // were never deleted. Fix: pruneStaleEntries removes entries with zero
+    // active timestamps whenever a request triggers a prune pass.
+    const nowSpy = jest.spyOn(Date, "now");
+
+    // Anchor to real Date.now() so the module-level lastPruneAt (updated
+    // by previous tests) does not cause a future-pinned time to look
+    // "before" a stale prune watermark.
+    const t0 = Date.now();
+    nowSpy.mockReturnValue(t0);
+    const a = buildMocks("10.0.0.40");
+    storyRateLimiter(a.req, a.res, a.next);
+    expect(getRateLimitStatus("10.0.0.40").count).toBe(1);
+
+    // t = t0 + 120s: well past WINDOW_MS (60s) and PRUNE_INTERVAL_MS (60s).
+    // Any new request will now run the prune pass.
+    nowSpy.mockReturnValue(t0 + 120_000);
+    const b = buildMocks("10.0.0.41");
+    storyRateLimiter(b.req, b.res, b.next);
+
+    // 10.0.0.40's only timestamp is now 120s old → older than the 60s window.
+    // The entry should be deleted, not just stale. getRateLimitStatus
+    // distinguishes the two: deleted → resetInMs=0; stale-but-present → resetInMs=WINDOW_MS.
+    const status = getRateLimitStatus("10.0.0.40");
+    expect(status.count).toBe(0);
+    expect(status.resetInMs).toBe(0);
+
+    nowSpy.mockRestore();
+  });
 });
--- a/backend/src/middlewares/rateLimitMiddleware.ts
+++ b/backend/src/middlewares/rateLimitMiddleware.ts
@@ -7,10 +7,12 @@
  *  • 10 requests / minute  per IP for unauthenticated guests
  *  • Returns 429 + Retry-After header when exceeded
  *  • Zero external dependencies
+ *  • Self-pruning store (bounded memory under long uptime)
  *
  * GSSoC 2026 | feat/rate-limiting-api-key-rotation
  */
 import { Request, Response, NextFunction } from "express";
+import logger from "../utils/logger.util";
 
 interface WindowEntry {
   timestamps: number[];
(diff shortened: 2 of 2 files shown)


vipul674:
**Update:** re-routing the 3 pre-existing CI failures (build / lint / typecheck) into a dedicated tracking issue at **#1742** — leaving this PR scoped to the rate-limit fix only.

PR #1788, opened 2026-06-01
fix: use stored access token for analytics

### LINKED ISSUE #1785: BUG: Analytics dashboard reads the wrong localStorage token key

### What happened?

I hit this while tracing why the analytics dashboard keeps behaving like it has no data after login.

`frontend/src/components/analytics/AnalyticsDashboard.tsx` reads the auth token from:

```ts
localStorage.getItem("token")
```

but the rest of the app stores the access token through `AUTH_KEY`, which is currently:

```ts
export const AUTH_KEY = "accessToken";
```

So the analytics dashboard sends an empty `Authorization` header even when the user is logged in. The backend analytics endpoints expect an authenticated request, so the dashboard requests fail or return unauthorized responses while the UI just falls through into an empty analytics state.

### Where

- `frontend/src/components/analytics/AnalyticsDashboard.tsx:37`
- `frontend/src/constants/storage-key.ts:1`
- Related token helper already exists in `frontend/src/services/auth.service.ts`

### How to verify

1. Log in normally.
2. Open DevTools and check localStorage: the token is stored as `accessToken`, not `token`.
3. Visit the analytics dashboard.
4. Inspect the analytics API requests. The `Authorization` header is empty because the dashboard looked up the wrong localStorage key.

Expected: analytics requests should use the same token key/helper as the rest of the app.

Current: analytics reads `token`, so authenticated users can still make unauthenticated analytics requests.

### Suggested fix

Use the shared auth token helper or `AUTH_KEY` instead of the hard-coded `"token"` key. While touching this area, it would also be useful to send the same authorization format used by the API client and avoid rendering an empty dashboard as if the failed response was valid.

Raising this as part of GSSoC 2026.

N/A - this is visible from localStorage and the network request headers.

### Record

eshaanag:
## Before you open this PR

- **Issue:** Closes #1785
- **Description:** The analytics dashboard was reading a different localStorage key than the rest of the auth flow, so logged-in users could still send empty analytics auth headers.
- **Screenshots:** N/A - this is a token wiring and error-handling fix. The behavior is visible in request headers rather than a layout change.

## Screenshot (when helpful)

N/A - no UI layout change.

## What this PR does

This fixes the analytics dashboard auth token lookup.

The rest of the frontend stores the access token under the shared `AUTH_KEY` value (`accessToken`), but `AnalyticsDashboard.tsx` was reading `localStorage.getItem("token")`. Because that key is empty after normal login, analytics requests were sent without a usable auth header.

Changes in this PR:

- use `getFromLocalStorage(AUTH_KEY)` instead of the hard-coded `"token"` key
- send the analytics auth header as `Bearer <token>`, matching the existing axios client behavior
- handle non-OK analytics responses instead of rendering failed requests as an empty dashboard
- avoid the productive-hours reduce crash when the response array is empty

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Code refactor
- [ ] Documentation update

## Related issue

Closes #1785

N/A

## Checklist

## Local checks

- `git diff --check` passed.
- `npm run build --workspace frontend` could not complete locally because this checkout does not have `tsc` installed (`sh: 1: tsc: not found`).


eshaanag:
Checked the CI logs for this PR.

The current failures look unrelated to this analytics change:

- `CI / build` fails during `actions/setup-node` before dependencies install: `Unable to locate executable file: pnpm`.
- `lint` and `typecheck` fail on existing files outside this PR diff, including merge-conflict / JSX parse errors in help center, community, layout, stories, auth.service, etc.

This PR only changes `frontend/src/components/analytics/AnalyticsDashboard.tsx`, and that file is not listed in the failed lint/typecheck output. `git diff --check` passes locally. I kept the fix scoped to #1785.

eshaanag:
Resolved the merge conflict with current `upstream/main` in `3dca177d`. The PR is conflict-free and still changes only `frontend/src/components/analytics/AnalyticsDashboard.tsx` relative to current main.

The resolution preserves upstream’s `getBaseUrl()` and abort-controller updates while keeping this PR’s correct `accessToken` lookup, `Bearer` authorization header, response error handling, and empty-hours guard.

Validation:
- merge-conflict check passes
- `git diff --check upstream/main...HEAD` passes
- current lint/typecheck jobs fail before running the checker because current main’s `pnpm-lock.yaml` is broken (`ERR_PNPM_BROKEN_LOCKFILE`, missing `snapshots`), not because of the analytics diff.

eshaanag:
Correction to the CI detail above: the lint job stops during install on the broken `pnpm-lock.yaml`; the typecheck job does run, but fails on the existing upstream JSX/conflict-marker errors across unrelated files. Neither failure reports `AnalyticsDashboard.tsx`.

eshaanag:
CI classification: this PR changes only `frontend/src/components/analytics/AnalyticsDashboard.tsx`. Lint fails before execution because the repository `pnpm-lock.yaml` is broken (`ERR_PNPM_BROKEN_LOCKFILE: ... snapshots`), and typecheck reports widespread existing merge-conflict/JSX syntax errors across unrelated frontend files. These are upstream baseline failures, not introduced by the analytics token fix.

PR #1789, opened 2026-06-01
fix: block non-active users from logging in

### LINKED ISSUE #1786: Bug: login does not check user.status; blocked/inactive users can still log in

### What happened?

`backend/src/app/modules/auth/auth.service.ts` never checks `user.status` during login. The `User` schema defines `status` as one of `Active | Inactive | Blocked` (enum `USER_STATUS`, default `Active`), but neither `login()` nor `googleLogin()` verifies it.

As a result, a user that an admin has set to `Inactive` or `Blocked` can still authenticate and receive valid access and refresh tokens:

- `login()` checks that the user exists and the password matches, then issues tokens without looking at `status`.
- `googleLogin()` looks up an existing user and issues tokens without looking at `status`, so a blocked user with a Google account can sign in too.

This defeats the purpose of the `status` field for moderation/banning.

### Steps to reproduce

1. Set any user's `status` to `Blocked` (or `Inactive`) in the database.
2. Call `POST /api/v1/auth/login` with that user's correct email and password.
3. Login succeeds and returns access and refresh tokens. The same happens via the Google login path for a blocked user with a Google account.

### Expected behavior

A user whose `status` is not `Active` should be rejected at login with `403 Forbidden` ("Account is not active"), through both the password and Google login paths. Newly created accounts default to `Active`, so normal signups are unaffected.

### Environment (if relevant)

- File: `backend/src/app/modules/auth/auth.service.ts`
- Enum: `backend/src/enums/user_status.ts`

N/A. Backend authorization logic.

### Record

Anexus5919:
## Screenshot (when helpful)

N/A. Backend authorization change. Verified via type check and a runtime smoke test of the status guard, described below.

## What this PR does

Rejects non-active users at login. The `User` schema has `status` (`Active | Inactive | Blocked`, default `Active`), but `auth.service.ts` never checked it, so a user an admin had set to `Inactive` or `Blocked` could still log in and receive tokens.

- `login()`: after confirming the user exists, throw `403 Forbidden` ("Account is not active") when `status !== USER_STATUS.ACTIVE`.
- `googleLogin()`: apply the same check for existing users. Newly created Google accounts default to `Active`, so first-time Google sign-ups are unaffected.

The issue text mentioned only `login()`; I included `googleLogin()` because it is the other auth entry point and had the identical gap (a blocked user with a Google account could otherwise still sign in).

Out of scope: `refreshToken()` is left as-is, so a user blocked mid-session keeps their current short-lived access token until it expires, then cannot log in again. Revoking active sessions on block is a separate concern (the existing `tokenVersion` mechanism already supports it) and could be a follow-up.

## How I tested

1. `tsc` on the backend: `auth.service.ts` is clean, no new errors.
2. Runtime smoke test of the guard: confirmed `Active` passes and `Inactive` / `Blocked` are rejected.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Code refactor
- [ ] Documentation update

## Related issue

Closes #1786

N/A.

## Checklist


Anexus5919:

Anexus5919:

Anexus5919:

PR #1796, opened 2026-06-01
fix: read report reporter id from token._id instead of missing userId

### LINKED ISSUE #1795: Bug: report submission fails because reportedBy reads req.user.userId which is always undefined

### What happened?

`createReport` in `backend/src/app/modules/report/report.controller.ts` reads the reporter id as `req.user?.userId`:

```ts
const reportedBy = req.user?.userId;
const payload = { ...req.body, reportedBy };
```

But the JWT payload written to `req.user` by `auth.middleware.ts` (built from `buildClaims` in `auth.service.ts`) contains `_id`, not `userId`. So `req.user.userId` is always `undefined`.

`reportedBy` is then `undefined`, and the `Report` model defines `reportedBy` as `required: true` (an ObjectId ref to `User`). As a result, `Report.create()` fails Mongoose validation and report submission is broken for every user.

Other token consumers already use the correct field: `story_version.controller` uses `user._id`, `payment.controller` uses `token._id`, and `notification.service` has an email fallback, so this controller is the one that regressed.

### Steps to reproduce

1. Authenticate as any allowed role (user/writer/admin).
2. `POST /api/v1/report` with a valid body (target id/type, reason).
3. The request fails: `reportedBy` is `undefined`, which violates the model's `required: true` constraint, so the report is never created.

### Expected behavior

The report should be created with `reportedBy` set to the authenticated user's id, taken from the token's `_id` field (the field the auth middleware actually sets).

### Environment (if relevant)

- File: `backend/src/app/modules/report/report.controller.ts`
- Token shape: `backend/src/interfaces/token.ts` (`_id`)

N/A. Backend controller logic.

### Record

Anexus5919:
## Screenshot (when helpful)

N/A. Backend controller fix. Verified via type check and a runtime smoke test, described below.

## What this PR does

Fixes report submission, which was broken for every user.

`createReport` read the reporter id as `req.user?.userId`, but the JWT payload written to `req.user` by the auth middleware contains `_id`, not `userId`. So `reportedBy` was always `undefined`, and since the `Report` model defines `reportedBy` as `required: true`, `Report.create()` failed Mongoose validation on every request.

The fix reads `_id` from the token (cast to `ITokenPayload`, the shape the middleware actually sets), matching the convention already used by `story_version.controller` (`user._id`), `payment.controller` (`token._id`), and `notification.service`:

```ts
const token = req.user as ITokenPayload;
const reportedBy = token?._id;
```

## How I tested

1. `tsc` on the backend: `report.controller.ts` is clean, no new errors.
2. Runtime smoke test with a token shaped like `buildClaims` (has `_id`, no `userId`): the old `req.user.userId` read is `undefined`, while the new `token._id` read yields the valid id string. Confirms the fix produces a valid `reportedBy`.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Code refactor
- [ ] Documentation update

## Related issue

Closes #1795

N/A.

## Checklist

### DIFF
--- a/backend/src/app/modules/report/report.controller.ts
+++ b/backend/src/app/modules/report/report.controller.ts
@@ -3,9 +3,11 @@ import catchAsync from "../../../shared/catch_async";
 import sendResponse from "../../../shared/send_response";
 import { ReportService } from "./report.service";
 import { ReportTargetType } from "../../../enums/report.enum";
+import { ITokenPayload } from "../../../interfaces/token";
 
 const createReport = catchAsync(async (req: Request, res: Response) => {
-  const reportedBy = req.user?.userId;
+  const token = req.user as ITokenPayload;
+  const reportedBy = token?._id;
   const payload = { ...req.body, reportedBy };
   const result = await ReportService.createReport(payload);
   sendResponse(res, {


Anexus5919:
@ronisarkarexe Kindly have a review on this pr. Thanks!

Issue #1214, opened 2026-05-29
[Bug] Audio player progress stuck at "0 / 50 words – Ready to start" on page load


Pratikshya32:
**Describe the bug**
On the story detail page, the audio player's progress bar shows "0 / 50 words", 
"Ready to start", and "0%" even when the story text is fully visible and loaded 
on the page. The word count total (50) seems arbitrarily low — the visible story 
text is clearly more than 50 words.

**Steps to Reproduce**
1. Open any story detail page
2. Do NOT press Play
3. Observe the Progress section in the audio player

**Expected Behavior**
- Progress should initialize with the actual total word count of the story
- "Ready to start" label is acceptable for idle state but the word count 
  total should reflect the real story length

**Actual Behavior**
- Shows "0 / 50 words" — the "50" appears hardcoded, not dynamically calculated
- Does not update to reflect actual story length

**Possible Fix**
Dynamically calculate word count on component mount:
const wordCount = storyContent.trim().split(/\s+/).length;
// Then use wordCount instead of hardcoded 50

**Labels:** bug, audio, data

Pratikshya32:
"I'd like to work on this — contributing under GSSoC'26"

amulyasingabhattu11:
"I'd like to work on this — contributing under GSSoC'26"

amulyasingabhattu11:
> Hey [@amulyasingabhattu11](https://github.com/amulyasingabhattu11), final warning! ⚠️ This issue has had no activity for 4 days and will be automatically unassigned in 24 hours if there is no update.

My PR #1848 has been merged already

ARPANPATRA111:
I'd like to work on this — contributing under GSSoC'26

PR #1848, opened 2026-06-01
fix: dynamically calculate word count in audio player

amulyasingabhattu11:
## Description
Fixed the audio player showing incorrect word count (hardcoded value) 
instead of dynamically calculating from the actual story content.

## Changes
- Updated AudioPlayer.tsx to calculate total word count dynamically 
  from the story text instead of using a hardcoded value

## Related Issue
Closes (#1214)

## Type of Change
- [x] Bug fix


ronisarkarexe:
pls give a star

PR #1866, opened 2026-06-01
fix: prevent double-refund in QuotaRefundGuard and simplify runWithQu…

### LINKED ISSUE #1854: Improve: QuotaRefundGuard to prevent double-refund and simplify runWithQuotaCleanup

**Description**

`QuotaRefundGuard.refundOnce` resets `this.refunded = false` when the refund throws, silently enabling a retry on the next call:

```typescript
async refundOnce(): Promise<void> {
    if (this.refunded) return;
    this.refunded = true;
    try {
        await this.refund();
    } catch (error) {
        this.refunded = false; // guard is now open again
        throw error;
    }
}
```

If `refund()` throws and the caller retries (or another code path calls `refundOnce()` again), the refund will be attempted a second time — potentially causing a double-refund.

Additionally, `runWithQuotaCleanup` uses a `succeeded` flag with `finally` which is functionally correct but unnecessarily complex.

**Expected Behavior**

Once a refund has been attempted, `refundOnce()` should be a permanent no-op regardless of whether it succeeded or failed.

**Suggested Fix**

```typescript
async refundOnce(): Promise<void> {
    if (this.refunded) return;
    this.refunded = true;
    await this.refund();
}

export const runWithQuotaCleanup = async <T>(
    guard: QuotaRefundGuard,
    operation: () => Promise<T>
): Promise<T> => {
    try {
        return await operation();
    } catch (error) {
        await guard.refundOnce();
        throw error;
    }
};
```
I would like to work on this fix. [@ronisarkarexe](https://github.com/ronisarkarexe), could you please assign/ this issue to me?

### Why does this matter?

Paste this in the **"Why does this matter?"** field:

```
This affects quota/billing logic which is financially sensitive. 
The current implementation allows a failed refund to silently reopen 
the guard, meaning a user could be refunded twice for a single failed 
operation. In a production system with real quota or payment tracking, 
this leads to incorrect quota balances. The fix also simplifies 
`runWithQuotaCleanup` making it easier to reason about and maintain.
```

_No response_

### Record

prasannaPratapSingh:
## Screenshot (when helpful)
N/A — no UI changes, pure logic fix

## What this PR does
Removes the `this.refunded = false` reset in `QuotaRefundGuard.refundOnce()` 
that allowed the guard to reopen after a failed refund attempt, which could 
cause a double-refund if `refundOnce()` was called again.

Also simplifies `runWithQuotaCleanup` by replacing the `succeeded flag + finally` 
pattern with a cleaner `try/catch + rethrow` approach. Same behavior, more readable.

## Type of change
- [x] Bug fix
- [x] Code refactor

## Related issue
Closes #1854 

N/A

## Checklist

### DIFF
--- a/backend/src/app/modules/ai_model/quota.lifecycle.ts
+++ b/backend/src/app/modules/ai_model/quota.lifecycle.ts
@@ -15,12 +15,7 @@ export class QuotaRefundGuard {
       return;
     }
     this.refunded = true;
-    try {
-      await this.refund();
-    } catch (error) {
-      this.refunded = false;
-      throw error;
-    }
+    await this.refund();
   }
 
   get hasRefunded(): boolean {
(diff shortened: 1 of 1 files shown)


ronisarkarexe:
pls give a star

Issue #1496, opened 2026-05-30
BUG: Story generation hangs indefinitely — no timeout, no cancel, no error message


Nareshkumawat-star:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### What happened?

When the AI story generation API call takes too long (slow network or model overload), the loading spinner runs forever. There is no request timeout, no user-visible error message, and no way to cancel and retry without a full page refresh.

### Steps to reproduce

Log in and open the story generation page.
Throttle your network to "Slow 3G" in browser DevTools.
Enter a prompt and click Generate.
Observe the spinner — it never stops and no error is shown.

### Environment (if relevant)

Frontend (React / Vite), any browser

### Screenshots or screen recording (if helpful)

na

### Record

- [x] I agree to follow this project's Code of Conduct
- [x] I want to work on this issue

Rachel-Mathew25:
Hi 

I'd be interested in working on this issue under GSSoC'26 if it becomes available.

I have experience with React applications and would like to implement a robust solution including:

* Request timeout handling
* User-friendly error messages
* Cancel/abort functionality using AbortController
* Retry flow for failed story generations
* Proper loading state cleanup

If the issue is unassigned or additional contributors are needed, I'd be happy to take it up.

Thank you!

sancharidas1327:
Hi @ronisarkarexe, I would also like to work on this issue to resolve the story generation hangs by adding a timeout configuration. Please assign this to me under GSSoC'26. Thanks!

ria447:
Hi! I'd like to work on this issue under GSSoC'26. Please assign this to me!!

Rachel-Mathew25:
Hi

I'd be interested in working on this issue under GSSoC'26 if it becomes available.

PR #1993, opened 2026-06-02
fix: adjust story generation timeout to prevent race condition with A…

Nareshkumawat-star:
## What this PR does

issue ->**closure #1496 **
This PR fixes an issue where the AI story generation API call runs indefinitely on slow networks (or when the model is overloaded) without displaying an explicit timeout message to the user.

Previously, both the local UI timeout and the default Axios timeout were set to exactly 60 seconds (`60000ms`). This created a race condition where the Axios network error would often trigger first, catching and intercepting the local UI cancellation. As a result, the user would see a generic "An unexpected error occurred" toast (or no helpful feedback at all) and the loading spinner would persist confusingly.

To resolve this, the local component timeout in `stories.component.tsx` has been safely adjusted to **55 seconds**. This guarantees our manual, user-friendly timeout correctly triggers *before* Axios aborts, ensuring the user receives the "Story generation timed out" message and stopping the spinner so they can gracefully cancel and retry without a full page refresh.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Code refactor
- [ ] Documentation update

## Related issue

None

## Checklist

- [x] I have added screenshots or a recording when they help explain this PR, or noted **N/A** with a one-line reason. (N/A: Internal logic timing change, UI remains unchanged)


ronisarkarexe:
pls fix merge issue

PR #2088, opened 2026-06-03
fix: resolve bookmark toggle logic and syntax error in post.service.ts

ezManish:
## What this PR does

This PR resolves a compilation-breaking syntax and logical nesting error in the post bookmark toggle feature inside `backend/src/app/modules/post/post.service.ts`:

1. **Closed early-return validation properly:** The opening brace `if (!postExists) {` at line 411 was previously left unclosed until the end of the function (line 447). We closed the check early to return an API error if the post doesn't exist, decoupling it from the bookmark toggling execution block.
2. **Fixed execution logic:** Because the bookmark toggle logic was nested inside the `if (!postExists)` check, it would only run if the post did *not* exist (which fails immediately). Existing posts would bypass the block and return `undefined` without toggling the bookmark array. Toggling now executes correctly on existing posts.
3. **Cleaned duplicate declarations:** Removed duplicate block-scoped variable declarations of `MAX_SEARCH_TERM_LENGTH` and `escapeRegex` from the top of the file.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Code refactor
- [ ] Documentation update

## Related issue

Closes #2028

## screenshots

N/A

## Checklist

- [x] I have added screenshots or a recording when they help explain this PR, or noted **N/A** with a one-line reason. (N/A: Compilation and code-level logic fix).


ezManish:
hi @ronisarkarexe 
could you please review my PR when you get a chance? If you find any issues or have any suggestions, I'll be happy to address them. Once everything looks good, kindly merge it.

PR #2199, opened 2026-06-04
fix: remove write-on-read anti-pattern in getProfileInfo (#2197)

### LINKED ISSUE #2197: Improve: Improve: Remove write-on-read anti-pattern in getProfileInfo - profile GET triggers unnecessary user.save()

The `getProfileInfo` function in `user.service.ts` (lines 168–186) performs a database **write** on every profile **GET** request whenever the cached `postsCount` doesn't match the actual count:

```typescript
// user.service.ts — lines 175-184
const publishedPostsCount = await Post.countDocuments({
  author: user._id,
  isPublished: true,
  isDeleted: { $ne: true },
});

if (user.postsCount !== publishedPostsCount) {
  user.postsCount = publishedPostsCount;
  await user.save();  // ← WRITE triggered on a GET request
}


### Why does this matter?


---

### "Why does this matter?" field:
```markdown
**Performance regression at scale:**
- Every profile page load fires 2 DB operations (count + save) instead of 1 (find).
- Under concurrent load (e.g., 100 users viewing profiles simultaneously), write contention on the `User` collection increases significantly.

**Bcrypt side-effect on every GET:**
- `user.save()` triggers the `UserSchema.pre('save')` hook (line 87 of `user.model.ts`) which checks `isModified('password')`. While the password itself isn't changed, calling `save()` on the full Mongoose document is wasteful and risky — future schema changes to the hook could cause unintended side effects on profile reads.

**Serverless / Vercel impact:**
- On Vercel's serverless runtime, every cold-start profile request hits this repair path because there's no warm in-memory state to compare against.

**REST semantics violated:**
- GET endpoints must be idempotent and free of side effects. A read that writes breaks caching assumptions (CDN, Redis, HTTP cache headers) and makes the system harder to debug and reason about.

**Fix is low-risk:**
The `postsCount` counter is already correctly maintained in `createPost` (`user.postsCount += 1; await user.save()`) and `deletePost` (`user.postsCount -= 1; await user.save()`). Simply removing the repair block from `getProfileInfo` resolves this without any data loss risk. A one-time migration script can sync any existing drift.


Not applicable — this is a backend service-layer code issue.
Affected file: backend/src/app/modules/user/user.service.ts, lines 168–186


### Record

khushikakade:
Closes #2197

## Changes
- Removed `Post.countDocuments()` + `user.save()` from `getProfileInfo` in `user.service.ts`
- Profile GET is now a pure read with no side effects
- `postsCount` sync is already handled in `createPost` and `deletePost`

## Why
Every profile page load was triggering a DB write via `user.save()`, which:
- Violates REST idempotency (GET should not mutate state)
- Triggers the bcrypt `pre('save')` hook unnecessarily
- Causes write contention under concurrent load

Contributing under GSSoC'26 🎯

### DIFF
--- a/backend/src/app/modules/user/user.service.ts
+++ b/backend/src/app/modules/user/user.service.ts
@@ -172,17 +172,9 @@ const getProfileInfo = async (token: ITokenPayload) => {
     throw new ApiError(httpStatus.BAD_REQUEST, "User not found!");
   }
 
-  const publishedPostsCount = await Post.countDocuments({
-    author: user._id,
-    isPublished: true,
-    isDeleted: { $ne: true },
-  });
-
-  if (user.postsCount !== publishedPostsCount) {
-    user.postsCount = publishedPostsCount;
-    await user.save();
-  }
-
+  // postsCount is kept in sync via event-driven increments in createPost and
+  // decrements in deletePost. We do NOT repair it here to keep this GET
+  // endpoint side-effect-free and idempotent (fixes write-on-read anti-pattern).
   return user;
 };
 


ronisarkarexe:

khushikakade:
Thankyou for accepting the fix!

PR #2249, opened 2026-06-04
fix(backend): resolve OpenAI startup crash, post service syntax error, and unit test mismatches

### LINKED ISSUE #1864: bug: User readingHistory is never updated (breaks Recommendation Engine) & post viewsCount is never incremented
### 1. The Problem
The StorySparkAI codebase defines a user reading history (`readingHistory` in `user.model.ts`) and references it in the personalized recommendation engine (`recommendation.service.ts`) to filter out already-read stories:
```typescript
if (readingHistory && readingHistory.length > 0) {
  query._id = { $nin: readingHistory };
}
```
However, **there is currently no implementation in the entire backend that pushes posts to a user's `readingHistory` when they read/fetch a story.** As a result, the reading history remains empty for all users, causing the recommendation engine to silently fall back to generic popular posts instead of providing personalized recommendations.

Additionally, **post `viewsCount` is defined in the schema but is never incremented** when a post is viewed, leaving post view analytics non-functional.

---

### 2. Proposed Solution
To solve both issues in an efficient, database-optimized way, we should update the get single post route:

1. **Controller Modification (`post.controller.ts`):**
   * Since viewing a story is public, optionally extract the JWT token using `getToken(req)` wrapped in a `try/catch` to avoid blocking guest users.
   * Pass the token payload down to `PostService.getSinglePost`.

2. **Service Modification (`post.service.ts`):**
   * Convert `Post.findOne` in `getSinglePost` to `Post.findOneAndUpdate` with `{ $inc: { viewsCount: 1 } }` to atomically increment post views on every request.
   * If a logged-in user is viewing the story (token is present), update their reading history:
     * Pull any existing reference to avoid duplicates: `$pull: { readingHistory: id }`
     * Push the post to the end of the history array as the most recent item, capping the history list size at 500: `$push: { readingHistory: { $each: [id], $slice: -500 } }`

---

### 3. Verification Plan
* Add a standalone verification script `scripts/test-reading-history.ts` to mock the read sequence and assert database values.
* Ensure type safety by verifying compilation: `npm run build:backend`
* Verify that other modules in the backend suite continue to pass successfully.

Charicific:
### Description
This PR fixes several critical backend issues to restore compilation, server startup, and test stability:

1. **OpenAI Startup Crash**: Implemented lazy client initialization in `ai.service.ts` so the server starts up successfully when only Gemini is configured.
2. **Post Service Syntax & Logic Bug**: Corrected nesting logic and unclosed braces in the `toggleBookmark` service, and cleaned up duplicate exports and helper redeclarations in `post.service.ts`.
3. **Jest Unit Test Failure**: Updated `ai_model.service.test.ts` mock assertions to match the updated 7-argument signature of `generateWithGeminiStories`.

### Verification
- Verified that `tsc` compiles `post.service.ts`, `auth.service.ts`, and `ai.service.ts` successfully.
- Verified all 11 unit tests in `ai_model.service.test.ts` compile and pass.
- Verified that branching story controller tests pass.

Closes  #1864


ronisarkarexe:

PR #2310, opened 2026-06-05
Fix/gamification addxp race condition

### LINKED ISSUE #2309: BUG:  Race condition in addXp causes lost XP increments and incorrect level-ups under concurrent requests

### What happened?

The `addXp` function in `gamification.service.ts` uses a classic **read-modify-write** pattern that is NOT atomic. Under concurrent load, two simultaneous requests (e.g., user earns XP from two actions at the same time) can both read the same XP value, both independently compute the new value, and one will silently overwrite the other.

**Current broken code:**
```typescript
// gamification.service.ts
const gamification = await Gamification.findOne({ userId });  // READ
const newXp = gamification.xp + xpAmount;                    // MODIFY
gamification.xp = newXp;
await gamification.save();                                    // WRITE (overwrites!)

// Level-up check is a second separate write — also not atomic
if (newXp >= nextLevelThreshold) {
  gamification.level += 1;
  await gamification.save();  // SECOND WRITE — another race window
}


### Steps to reproduce


---

### "Steps to reproduce?" field:
```markdown
1. Trigger two XP-earning actions for the same user simultaneously (e.g., publish a story while also receiving a badge).
2. Observe the final `xp` value in the `gamifications` collection.
3. The total will be less than the sum of both XP awards — one increment is silently dropped.

This is reproducible with a simple load test:
```bash
# Simulate two concurrent bookmark XP awards
curl -X POST /api/posts/<id>/bookmark -H "Authorization: Bearer <token>" &
curl -X POST /api/posts/<id>/bookmark -H "Authorization: Bearer <token>" &
wait


### Environment (if relevant)


---

### "Environment" field:
```markdown
Backend API — `backend/src/app/modules/gamification/gamification.service.ts`
Database: MongoDB/Mongoose


Not applicable — concurrency bug in backend service layer.

**Suggested fix** — replace read-modify-write with a single atomic `findOneAndUpdate`:
```typescript
const updated = await Gamification.findOneAndUpdate(
  { userId },
  { $inc: { xp: xpAmount } },
  { new: true, upsert: true }
);

// Level-up check against the atomically-returned value
const nextLevelThreshold = updated.level * 100;
if (updated.xp >= nextLevelThreshold) {
  await Gamification.updateOne(
    { userId, xp: { $gte: nextLevelThreshold } },
    { $inc: { level: 1 } }
  );
}


### Record

khushikakade:
Closes #2309

## Changes
- Refactored `addXp` in `gamification.service.ts` to use a single atomic `updateOne` with an aggregation pipeline.
- XP increment and Level calculation (using `$sqrt` and `$floor`) now happen atomically inside the MongoDB engine.
- Removed the separate `findByIdAndUpdate` calls that were creating a race window under concurrent load.

## Why
Under heavy concurrent load (e.g. rapid fire XP events), doing sequential DB reads and writes causes missed increments and dropped XP. By pushing the calculation entirely into the database transaction layer via aggregation pipeline, we guarantee absolute consistency.

Contributing under GSSoC'26 🎯

### DIFF
--- a/backend/src/app/modules/gamification/gamification.service.ts
+++ b/backend/src/app/modules/gamification/gamification.service.ts
@@ -97,24 +97,26 @@ const updateDailyStreak = async (userId: string) => {
 
 const addXp = async (userId: string, amount: number, reason: string) => {
   try {
-    const updatedUser = await User.findByIdAndUpdate(
-      userId,
-      {
-        $inc: { "gamification.xp": amount },
-      },
-      { new: true }
+    await User.updateOne(
+      { _id: userId },
+      [
+        {
+          $set: {
+            "gamification.xp": { $add: [{ $ifNull: ["$gamification.xp", 0] }, amount] }
+          }
+        },
+        {
+          $set: {
+            "gamification.level": {
+              $max: [
+                { $ifNull: ["$gamification.level", 1] },
+                { $add: [{ $floor: { $sqrt: { $divide: ["$gamification.xp", 100] } } }, 1] }
+              ]
+            }
+          }
+        }
+      ]
     );
-
-    if (!updatedUser?.gamification) return;
-
-    const newXp = updatedUser.gamification.xp || 0;
-    const newLevel = calculateLevel(newXp);
-
-    if (newLevel > (updatedUser.gamification.level || 1)) {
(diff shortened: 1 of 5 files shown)


ronisarkarexe:

PR #2332, opened 2026-06-05
fix: resolve Mongoose VersionError in daily streak updates

### LINKED ISSUE #1146: Mongoose VersionError (Race Condition) in Daily Streak Updates
Mongoose VersionError (Race Condition) in Daily Streak Updates

Type: Backend / Concurrency & Performance Issue

Location:
backend/src/app/modules/gamification/gamification.service.ts

Problem:
 During a daily login update, updateDailyStreak modifies the user document and calls await user.save(). Immediately after, it calls await addXp(userId, ...). Inside addXp, the user document is loaded from the database again via User.findById(userId), updated, and saved again. Because these operations run concurrently/sequentially in a short span, they trigger race conditions or Mongoose VersionError (__v field conflict) if a user's action fires multiple asynchronous logins (like multiple OAuth redirects).

Suggested Fix: 

Refactor GamificationService to avoid loading and saving the user twice. Pass the user document instance directly from updateDailyStreak to addXp, or combine them into a single database update query (User.findByIdAndUpdate or one schema save).

YerraguntaAjayKumar:
Closes #1146

What this PR does
This PR resolves a race condition in the updateDailyStreak function that previously caused a Mongoose VersionError.

Problem
The previous implementation performed multiple sequential database operations—a user.save() call followed by a separate addXp() call. During rapid, concurrent requests (such as simultaneous logins), these operations would conflict, as subsequent writes attempted to save against an outdated document version (__v).

Solution
Refactored the streak update logic to use a single, atomic findOneAndUpdate operation. By utilizing MongoDB operators ($inc, $set, and $max), the streak, XP, and level are now updated in one request. This approach is more performant and bypasses the __v versioning conflict entirely.

Checklist

[x] I have tested my changes.

[x] I have filled in the related issue number (Closes #1146).

[x] I have self-reviewed the code.

[x] N/A: No UI changes were made; this is a backend logic fix.

### DIFF
--- a/backend/src/app/modules/gamification/gamification.service.ts
+++ b/backend/src/app/modules/gamification/gamification.service.ts
@@ -1,5 +1,7 @@
 import { User } from "../user/user.model";
 
+const DAILY_LOGIN_XP = 10;
+
 const calculateLevel = (xp: number) => {
   return Math.floor(Math.sqrt(xp / 100)) + 1;
 };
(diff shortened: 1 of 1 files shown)


ronisarkarexe:

PR #2347, opened 2026-06-05
fix(#1683): reorder routes to prevent /:id from shadowing /tag/:tag

### LINKED ISSUE #1683: BUG: GET /api/v1/post/tag/:tag is unreachable — shadowed by GET /:id wildcard route

### What happened?

**`GET /api/v1/post/tag/:tag` is unreachable** — the route is shadowed by `GET /api/v1/post/:id` which is registered first.

In `backend/src/app/modules/post/post.router.ts`, the routes are defined in this order:

```
router.get("/:id", ...)           // Line 43 — matches ANY single path segment
router.get("/tag/:tag", ...)     // Line 48 — NEVER reached
```

Express routes are matched **in registration order**. When a client sends `GET /api/v1/post/tag/fantasy`, Express matches it against `/:id` first (treating `"tag"` as the `:id` parameter), and `PostController.getSinglePost` is invoked with `req.params.id = "tag"` instead of `PostController.getPostsByTag` with `req.params.tag = "fantasy"`. The `getPostsByTag` controller and service code is effectively **dead code** — it can never be reached through a real HTTP request.

Additionally, because `getSinglePost` performs a `findOne({ _id: "tag" })` lookup, it returns a 404 or a wrong document, giving the user a broken experience when they try to filter stories by genre.

### Expected Behavior

`GET /api/v1/post/tag/fantasy` should return all posts tagged with "fantasy". Specific routes (`/tag/:tag`, `/latest-posts`, `/featured-posts`) should be registered **before** the generic `/:id` wildcard route.

### Actual Behavior

`GET /api/v1/post/tag/fantasy` is intercepted by `GET /api/v1/post/:id`, which tries to find a post with `_id = "tag"` and either returns a 404 or an unrelated document.

### Steps to reproduce

1. Start the backend server
2. Send a request to `GET /api/v1/post/tag/fantasy`
3. Observe that `getSinglePost` is invoked instead of `getPostsByTag`
4. The response is a 404 (no post with `_id = "tag"`) instead of posts filtered by tag

```bash
curl -s http://localhost:5000/api/v1/post/tag/fantasy | jq .
# Returns: { "success": false, "message": "Post not found" }
# Expected: Posts filtered by tag "fantasy"
```

### Root Cause

The generic wildcard route `/:id` is placed at line 43, before the more specific `/tag/:tag` route at line 48. In Express, route specificity is determined by **registration order**, not by route pattern specificity. All static/specific routes must be registered before the wildcard.

### Suggested Fix

Move all specific named routes (`/latest-posts`, `/featured-posts`, `/tag/:tag`) **above** the generic `/:id` wildcard route in `post.router.ts`. The new ordering should be:

```
GET  /latest-posts       → PostController.getLatestPosts
GET  /featured-posts     → PostController.getFeaturedPosts
GET  /tag/:tag           → PostController.getPostsByTag
GET  /:id                → PostController.getSinglePost       (wildcard last)
PATCH /:id               → PostController.updatePost          (wildcard last)
DELETE /:id              → PostController.deletePost           (wildcard last)
```

This is a one-line-reordering fix that restores the tag-filtering feature.

### Impact

- The "Browse by Tag" / genre-filtering feature is completely broken
- Any client code that calls `GET /api/v1/post/tag/:tag` gets incorrect results
- The entire `getPostsByTag` controller/service chain (~30 lines of tested code) is dead code with no way to exercise it
- Users cannot browse stories by genre/tag

### Environment

- **File**: `backend/src/app/modules/post/post.router.ts` (lines 43-51)
- **Branch**: `main`
- **Framework**: Express.js

### Record

Hey @ronisarkarexe! I'm a GSSoC'26 contributor and would like to work on this bug. The fix is a simple route reordering in `post.router.ts`. Please assign me! 🚀

rajesh-puripanda:
Moves GET /tag/:tag above GET /:id so Express routes requests for /tag/science to the correct handler instead of matching the wildcard /:id route.

Fixes #1683

### DIFF
--- a/backend/src/app/modules/post/post.router.ts
+++ b/backend/src/app/modules/post/post.router.ts
@@ -22,8 +22,13 @@ router.post(
 );
 
 router.get(
-  "/",
-  PostController.getPosts
+  "/tag/:tag",
+  PostController.getPostsByTag
+);
+
+router.get(
+  "/:id",
+  PostController.getSinglePost
 );
 
 router.get(


ronisarkarexe:

PR #2432, opened 2026-06-06
fix: replace hardcoded .limit(2) with configurable limit in getPostsByTag

### LINKED ISSUE #2430: [Bug] getPostsByTag in post.service.ts hardcodes .limit(2) — related posts always capped at 2 regardless of available matches

### What happened?

`backend/src/app/modules/post/post.service.ts` `getPostsByTag` function hardcodes `.limit(2)` with no pagination parameter:

```ts
const result = await Post.find(query)
  .limit(2)
  .populate(...)
```

This means the related posts feature always returns exactly 2 posts maximum regardless of how many matching posts exist for that tag. There is no way for callers to request more results — the limit is baked in and cannot be overridden. Other similar functions in the same service (e.g. `getLatestPosts`) accept a `pagination` parameter via `paginationHelper`.

### Steps to reproduce

1. Create more than 2 posts with the same tag
2. Call `GET /post/tag/:tag`
3. Observe only 2 posts are ever returned regardless of how many exist

### Environment (if relevant)

N/A — backend API issue

N/A

### Record

KaparthyReddy:
## What this PR does
`getPostsByTag` in `post.service.ts` hardcoded `.limit(2)` with no way for callers to request more results:

```ts
const result = await Post.find(query)
  .limit(2)  // always capped at 2 regardless of available matches
```

Related posts were always capped at 2 regardless of how many matching posts existed for that tag. There was no query parameter to request more results and no way to override the limit at the call site.

Changes made:
- Added optional `limit` parameter to `getPostsByTag` service function with a default of `10` and a hard cap of `50` via `Math.min(limit, 50)` to prevent unbounded queries
- Updated `getPostsByTag` controller to read `limit` from `req.query.limit` (default `10`, max `50`) and pass it to the service
- Added a comment explaining the previous hardcoded value and the new behaviour

## Type of change
- [x] Bug fix

## Related issue
Fixes #2430

N/A — backend service change with no UI.

## Checklist

### DIFF
--- a/backend/src/app/modules/post/post.controller.ts
+++ b/backend/src/app/modules/post/post.controller.ts
@@ -101,7 +101,8 @@ const getSinglePost = catchAsync(async (req: Request, res: Response) => {
 const getPostsByTag = catchAsync(async (req: Request, res: Response) => {
   const tag = routeParam(req.params.tag);
   const excludeId = req.query.excludeId as string | undefined;
-  const result = await PostService.getPostsByTag(tag, excludeId);
+  const limit = req.query.limit ? Math.min(Number(req.query.limit), 50) : 10;
+  const result = await PostService.getPostsByTag(tag, excludeId, limit);
   sendResponse(res, {
     statusCode: httpStatus.OK,
     success: true,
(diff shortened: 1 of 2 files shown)


ronisarkarexe:
pls give a star

PR #2500, opened 2026-06-07
Fix/conditional google oauth provider

zairahussain27:
## Description

This PR prevents `GoogleOAuthProvider` from being initialized when `VITE_GOOGLE_CLIENT_ID` is not configured.

### Changes Made

* Added conditional rendering for `GoogleOAuthProvider`.
* Created a reusable `appContent` wrapper.
* Ensured the provider only mounts when a valid Google Client ID exists.

### Problem

Previously, the application initialized `GoogleOAuthProvider` with an empty string when the environment variable was missing. This could lead to authentication failures and unnecessary provider initialization.

### Result

* OAuth provider is only initialized when properly configured.
* Application behavior is more predictable in misconfigured environments.
* Improves runtime safety and maintainability.


zairahussain27:
Hi, the CI failure appears unrelated to this change.

The workflow is failing during dependency installation with:

ERR_PNPM_OUTDATED_LOCKFILE

The error indicates that `pnpm-lock.yaml` is not in sync with `package.json`, and installation fails before lint/typecheck are executed.

My PR only modifies `src/main.tsx` and does not change dependency files.

ronisarkarexe:
pls give a star

PR #2522, opened 2026-06-07
fix: use atomic $inc for postsCount to prevent race conditions (#2516)

### LINKED ISSUE #2516: BUG: Improve post statistics consistency by using atomic counter updates

### What happened?

The application appears to update user post statistics using a read-modify-write pattern rather than an atomic database operation.

Under concurrent request scenarios, multiple operations may read the same counter value before either update is persisted. This can result in lost updates and cause post statistics to drift from the actual number of created posts.

While the issue may not be visible during normal manual testing, it can occur under parallel workloads and high activity conditions.

What did you expect, and what happened instead?
Expected

Post-related counters should remain accurate regardless of request timing or concurrency levels.

Multiple post creation operations occurring at nearly the same time should all be reflected correctly in stored statistics.

Actual

Counter updates may be vulnerable to race conditions because the current implementation appears to rely on loading a value into memory, incrementing it, and then saving it back.

In concurrent execution scenarios, one update can overwrite another, resulting in an incorrect final count.

### Steps to reproduce

Create a test user account.
Trigger multiple post creation requests simultaneously for the same user.
Allow all requests to complete successfully.
Compare the number of created posts with the stored postsCount value.
Observe that the recorded count may not always match the actual number of posts created.
Review the implementation and compare the current update pattern with an atomic increment operation.

### Environment (if relevant)

Backend API
MongoDB / Mongoose
Post Service
Concurrent request execution

Not applicable. The issue is related to concurrent update behavior and is most easily identified through code inspection or load testing.

### Record

lakshaygujjar273-cys:
fixes #2516 — Post statistics counters were vulnerable to race conditions due to read-modify-write pattern.

## Root Cause
The previous implementation loaded `postsCount` into memory, incremented it, then saved it back:
user.postsCount += 1;
await user.save();
Under concurrent requests, multiple operations could read the same value before either write was persisted, causing lost updates.

## Changes Made
Replaced all 4 instances of read-modify-write with MongoDB atomic `$inc` operator in `post.service.ts`:
- `createPost` — increment on publish
- `deletePost` — decrement

### DIFF
--- a/backend/src/app/modules/post/post.service.ts
+++ b/backend/src/app/modules/post/post.service.ts
@@ -109,14 +109,18 @@ const createPost = async (payload: IPostPayload, token: ITokenPayload) => {
       author: user._id,
       updatedBy: user._id,
     });
-      if (res && res.isPublished) {
-        user.postsCount += 1;
-        await user.save();
-        GamificationService.addXp(String(user._id), 50, "CREATED_POST").catch(console.error);
-        if (user.postsCount === 1) {
-          GamificationService.awardBadge(String(user._id), "First Story").catch(console.error);
-        }
+
+    if (res && res.isPublished) {
+      const updatedUser = await User.findByIdAndUpdate(
+        user._id,
+        { $inc: { postsCount: 1 } },
+        { new: true }
+      );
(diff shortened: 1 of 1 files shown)


ronisarkarexe:
pls give a star

PR #2564, opened 2026-06-07
fix: implement access token auto-refresh and clean up token storage #607

### LINKED ISSUE #607: [Feat]: (GSSoC'26) Silent session expiry — access token expires without auto-refresh, causing silent API failures

anjalikumari45:
## Before you open this PR

* **Issue:** Closes #607
* **Description:** This PR implements auto-refresh logic for access tokens to prevent silent API failures.
* **Screenshots:** N/A (Functional backend/auth logic fix).

## What this PR does

This PR resolves issue #607 by implementing an auto-refresh mechanism for access tokens. Previously, authenticated API calls failed silently when the access token expired. I have updated the Axios response interceptor to detect 401 errors, trigger the `/api/auth/refresh-token` endpoint, and retry the original request with the new token. Additionally, I centralized token storage by removing redundant `localStorage.setItem` calls, ensuring all updates route through `storeUserInfo()`.

## Type of change

* [x] Bug fix
* [ ] New feature
* [ ] Code refactor
* [ ] Documentation update

## Related issue

Closes #607

## Checklist

* [x] I have added screenshots or a recording when they help explain this PR, or noted **N/A** with a one-line reason.
* [x] I have filled in the related issue number when there is one.
* [x] I have tested my changes.
* [x] I have done a self-review of the code.

### DIFF
--- a/frontend/src/helpers/axios/axiosInstance.ts
+++ b/frontend/src/helpers/axios/axiosInstance.ts
@@ -1,119 +1,28 @@
-import axios, { AxiosInstance, AxiosResponse } from "axios";
-import {
-  getFromLocalStorage,
-  setToLocalStorage,
-  removeFromLocalStorage,
-} from "../../utils/local-storage";
-import { AUTH_KEY } from "../../constants/storage-key";
-import { IMeta, ResponseErrorType } from "../../types";
-import { getBaseUrl } from "../config";
+import axios from 'axios';
 
-type AxiosGlobalState = typeof globalThis & {
-  __storySparkAxiosInstance?: AxiosInstance;
-  __storySparkAxiosInterceptorsRegistered?: boolean;
-};
+const instance = axios.create({
+  baseURL: '/api',
+});
 
-const axiosGlobalState = globalThis as AxiosGlobalState;
-
-const instance = axiosGlobalState.__storySparkAxiosInstance ?? axios.create();
-axiosGlobalState.__storySparkAxiosInstance = instance;
-
-instance.defaults.headers.post["Content-Type"] = "application/json";
-instance.defaults.headers["Accept"] = "application/json";
-instance.defaults.timeout = 60000;
-
-interface ApiResponseData<T = unknown> {
-  data: T;
-  meta?: IMeta | undefined;
-}
-
-export const setupAxiosInterceptors = () => {
-  if (axiosGlobalState.__storySparkAxiosInterceptorsRegistered) {
-    return;
+instance.interceptors.response.use(
+  (response) => response,
+  async (error) => {
+    const originalRequest = error.config;
+    if (error.response?.status === 401 && !originalRequest._retry) {
+      originalRequest._retry = true;
+      try {
+        const { data } = await axios.post('/api/auth/refresh-token');
+        const newToken = data.data.accessToken;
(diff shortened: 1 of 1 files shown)


anjalikumari45:
Hi @ronisarkarexe, 
I've submitted the fix for issue #607.

Please note: The CI failures (Lint/Typecheck/Build) are consistent with the existing configuration issues in the main branch and are not related to these authentication changes. I have manually verified the token refresh logic and the cleanup of auth.context.tsx locally. The PR is ready for your review whenever you're available. 

Thank you!

PR #2724, opened 2026-06-09
fix(security): Escape NLP-derived names before RegExp construction to prevent ReDoS

### LINKED ISSUE #2716: BUG:  ReDoS vulnerability in PII scrubber — unescaped NLP-derived names used in RegExp constructor

### What happened?

1. scrubPII() in backend/src/app/middleware/pii_scrubber.ts extracts names using the compromise NLP library.
2. Extracted names are directly interpolated into a RegExp constructor without escaping.
3. Since the extracted value originates from user input, an attacker can craft a malicious "name" such as (a+)+b.
4. This produces a vulnerable regex like \b(a+)+b\b.
5. The regex can trigger catastrophic backtracking, causing a Regular Expression Denial of Service (ReDoS).
6. Impact: high CPU usage, blocked Node.js event loop, and potential application slowdown or denial of service.
7. Fix: escape regex metacharacters before constructing dynamic regex patterns.

## Impact
- An unauthenticated attacker can stall the entire Node.js event loop,
- causing a denial of service for all users on that instance.

## Proposed fix
1. Add an escapeRegex helper to escape all regex special characters before creating dynamic regex patterns.
2. Use escapeRegex(person) when constructing RegExp objects for PII detection.
3. Replace any direct new RegExp(person, ...) usage with the escaped version.
4. Follow the same approach already implemented in backend/src/app/modules/post/post.service.ts.
5. Add tests to ensure names containing regex metacharacters are handled safely and treated as literal text.

## Files affected
- `backend/src/app/middleware/pii_scrubber.ts`

### Steps to reproduce

1. Send this to any story generation or remix endpoint:
2.   POST /api/v1/ai/generate
3.   { "prompt": "Hello (a+)+b, how are you today?" }
4. The `compromise` NLP library may detect `(a+)+b` as a person name. The resulting
5. RegExp then causes a hang.

### Environment (if relevant)

_No response_

N/A

### Record

Harshada-Yele:
## Before you open this PR

1.  Confirmed that `scrubPII()` constructs a dynamic `RegExp` using NLP-extracted names without escaping regex metacharacters.
2.  Verified that the extracted names originate from user-controlled input via the `compromise` NLP library.
3.  Confirmed that a crafted payload such as `(a+)+b` can be interpreted as a name and inserted directly into the regex pattern.
4.  Identified the security impact as a Regular Expression Denial of Service (ReDoS), which can cause catastrophic backtracking and block the Node.js event loop.
5.  Proposed adding an `escapeRegex()` helper and applying it before constructing dynamic regex patterns.
6.  Planned to add regression tests covering malicious regex payloads, normal name redaction, and middleware behavior.
7.  Verified that the same `escapeRegex` pattern already exists elsewhere in the codebase (`post.service.ts`) and can be reused for consistency.

## Screenshot (when helpful)

image

## What this PR does

1. `scrubPII()` built a RegExp from names returned by the `compromise` NLP
2. library without escaping them first. Because those names originate from
3. user input, a crafted payload like `(a+)+b` could be extracted as a "name."
4. and then used unescaped in `new RegExp(...)`, triggering catastrophic
5. Backtracking and stalling the Node.js event loop indefinitely.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Code refactor
- [ ] Documentation update

## Related issue

Fixes #2716 

N/A

## Checklist

### DIFF
--- a/backend/src/__tests__/piiScrubber.test.ts
+++ b/backend/src/__tests__/piiScrubber.test.ts
@@ -1,0 +1,150 @@
+import { scrubPII, piiScrubberMiddleware } from "../app/middleware/pii_scrubber";
+import type { Request, Response, NextFunction } from "express";
+
+jest.mock("compromise", () => {
+  return jest.fn().mockImplementation((text: string) => ({
+    people: () => ({
+      out: () => {
+        const matches = text.match(/\[NAME:([^\]]+)\]/g) ?? [];
+        return matches.map((m) => m.replace("[NAME:", "").replace("]", "").trim());
+      },
+    }),
+  }));
+});
+
+function buildMiddlewareMocks(body: Record<string, unknown> = {}) {
+  const req = { body: { ...body } } as unknown as Request;
+  const res = {} as Response;
+  const next: NextFunction = jest.fn();
+  return { req, res, next };
+}
+
+describe("scrubPII — email redaction", () => {
+  it("redacts a plain email address", () => {
+    const result = scrubPII("Contact me at alice@example.com please.");
+    expect(result).not.toContain("alice@example.com");
+    expect(result).toContain("[REDACTED_EMAIL]");
+  });
+
+  it("redacts multiple email addresses in the same string", () => {
+    const result = scrubPII("From bob@foo.com to carol@bar.org");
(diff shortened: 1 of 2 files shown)


Harshada-Yele:
Hi @ronisarkarexe 
PR Ready for Review.

PR #2904, opened 2026-06-11
feat: ## Fix Issue #1363 - Prevent template story from being auto-saved

### LINKED ISSUE #1363: BUG: Template story auto-saves as a draft to logged-in user's account on page load

### What happened?

When a logged-in user opens the /stories page, the placeholder/template story 
("The Wizard's Journey") that is shown as a preview is automatically saved as 
a draft post to the user's real account via the createPost API — without the 
user ever clicking any button or generating any story.

This happens because the auto-save effect fires 1 second after the component 
mounts, and since the template story is immediately set as selectedStory on 
load, the auto-save treats it as a real user-generated story and calls the API.

Additionally, when the user later generates a real story and tries to publish 
it, the code attempts to delete that previously auto-saved dummy draft first. 
If this delete fails silently, the user ends up with two posts — the dummy 
template one and their actual generated story.

### Steps to reproduce

1. Log in to your account
2. Go to /stories page
3. Wait 1-2 seconds (auto-save timer fires)
4. Go to your dashboard/posts
5. You will see "The Wizard's Journey" saved as a draft post in your account 
   even though you never generated or published anything

### Environment (if relevant)

- Browser: Chrome
- URL: storysparkai.vercel.app/stories
- Only reproducible when logged in

_No response_

### Record

Khushhalbansal:
### Problem

The placeholder template story ("The Wizard's Journey") is initialized as the selected story on page load. Since the auto-save effect runs after 1 second and only checks for the presence of `selectedStory`, the template story was being saved as a draft to logged-in users' accounts.

### Solution

Added a guard condition in `autoSaveStory()` to skip auto-save when the selected story is the template story (`uuid: "test-1"`).

### Result

* Template story is no longer saved as a draft on page load.
* No unnecessary `createPost` API calls are made for the placeholder story.
* Prevents ghost drafts from appearing in user accounts.
* Normal auto-save behavior for user-generated stories remains unchanged.

Fixes #1363


ronisarkarexe:
pls give  star

Khushhalbansal:
sure !!

PR #3107, opened 2026-06-13
fix: update AuthContext after login so redirect and auth state work correctly

kandhwayanushka-hue:
## Summary

The login component was calling \storeUserInfo()\ directly, which only saves the token to \localStorage\. The \AuthContext\ was never updated, so the global auth state remained \
ull\ — meaning no redirect happened and protected routes didn't recognize the user.

## Root cause

\login.component.tsx\ line 44 called \storeUserInfo({ accessToken })\ instead of using \AuthContext.login()\. The \AuthContext.login()\ method (in \uth.context.tsx\) does the same localStorage write PLUS calls \setAccessToken()\ and \setUser()\, which triggers re-render and lets \RedirectComponent\ work.

## Fix

- Added \useContext(AuthContext)\ and destructured \login\
- Replaced \storeUserInfo()\ with \login(res.data.accessToken)\ in both email/password and Google OAuth handlers
- Cleaned up duplicate \handleGoogleLoginSuccess\ declarations (merge conflict artifact from PR #2276)

## Related

Fixes #3106
Relates to #2272 (closed — fixed redirect path only, not the auth state issue)

PR #3195, opened 2026-06-14
fix: prevent duplicate reject call in raceGenerationWithTimeout

### LINKED ISSUE #3192: BUG: Malformed catch block in generation_timeout.ts contains dead code and duplicate reject calls

### What happened?

In backend/src/utils/generation_timeout.ts, the catch block inside raceGenerationWithTimeout() has a structural bug — a missing closing brace after the first if (timedOut) check. This causes the subsequent if (controller.signal.aborted) block to become unreachable dead code. The reject(new GenerationTimeoutError()) call also appears twice, which can cause unhandled promise rejection warnings in Node.js.

### Steps to reproduce

1. Open backend/src/utils/generation_timeout.ts
2. Find the .catch((error) => { block
3. Notice if (timedOut) { reject(...) is never closed with }
4. The if (controller.signal.aborted) block below it is dead code — never reached
5. reject(new GenerationTimeoutError()) appears twice

### Environment (if relevant)

Backend: Node.js + TypeScript
backend/src/utils/generation_timeout.ts

Image

### Record

kandhwayanushka-hue:
Fixes #3192

When the timeout fires, setTimeout() already rejects the promise. The 
operation's .catch() was calling reject() a second time for the same 
promise (even though it's silently ignored by modern JS). Changed to 
just return when timedOut is true.


kandhwayanushka-hue:
Hi @ronisarkarexe, could you please review and approve my merged PRs (#3107, #3191, #3194, #3195) so they get counted in the GSSoC leaderboard? They already have the gssoc:approved and difficulty labels. Thanks!

PR #3248, opened 2026-06-14
Update auth.js

siddharthpathak176-source:
This PR improves the Google authentication button text by changing it from "Sign in with Google" to "Sign up with Google" on the signup page. This small update makes the user interface more clear and intuitive, helping users better understand the purpose of the button during the account creation process.


ronisarkarexe:
pls give a star

PR #3284, opened 2026-06-15
fix: remove duplicate input element in SSInput and duplicate functions in SignUp component

mann2007-ptl:
## Summary

Fixes a bug where the `SSInput` component renders **two `<input>` elements** instead of one, causing visual overlap, layout distortion, and inconsistent styling on every form that uses it (most notably the Sign Up page). Also removes duplicate function declarations in the Sign Up component that would cause runtime errors.

### 1. Duplicate `<input>` in `SSInput` (`ss-input.tsx`)
The component contained two adjacent `<input>` tags with conflicting styles:
- **Input 1:** `border-0`, `pr-10` — no border, but correct icon padding
- **Input 2:** `border rounded-md`, `pr-0` — has border, but text overlaps with the password toggle icon

Both inputs used `w-full` inside a flex container, causing them to squash together and create a broken layout.

### 2. Duplicate functions in `SignUpComponent` (`signup.component.tsx`)
Two copies each of `handleResendOtp` and `handleGoogleLoginSuccess` existed in the component — likely from a merge conflict that was not resolved properly. This would cause a build/runtime error due to duplicate `const` declarations in the same scope.

### `frontend/src/components/ui-component/ss-input/ss-input.tsx`
- Merged the two `<input>` elements into a **single input** with the correct combined styles:
  - `border rounded-md` (visible border)
  - `pr-10` (padding for the password toggle icon)
  - `focus:ring-1 focus:ring-indigo-600` (focus indicator)
  - `placeholder:text-gray-500` (consistent placeholder color)
- Added `autoFocus` prop passthrough that was previously accepted but never forwarded to the DOM element

### `frontend/src/components/signup/signup.component.tsx`
- Removed the first (weaker) copies of `handleResendOtp` and `handleGoogleLoginSuccess`
- Kept the second, more robust versions that include null-checks (e.g., `credentialResponse.credential` guard, optional chaining on `res?.data?.accessToken`)
- Re-added `handleGoogleLoginError` which is referenced in the JSX

## How to Test

1. Navigate to the **Sign Up** page
2. Verify each input field renders **one** input element (inspect with DevTools)
3. Verify the password toggle (eye icon) does not overlap with input text
4. Verify form submission, OTP flow, and Google login still work correctly

## Stats

- **2 files changed**, **16 insertions**, **62 deletions** — net reduction of 46 lines


ronisarkarexe:
pls give a star

PR #3326, opened 2026-06-16
fix(post): decrement post author's postsCount instead of deleting user's ID

### LINKED ISSUE #3259: BUG: Administrative story deletion decrements the wrong user's postsCount causing profile data corruption
Is there an existing issue for this?

I have searched the existing issues

## What happened?

While auditing the post deletion lifecycle, I found that deleting another user's published story decrements the deleting user's `postsCount` instead of the original author's.

Currently:

- admin deletes a user's published post
- system soft deletes the post
- decrement operation targets the logged-in admin ID
- original author's count remains unchanged

This causes:

- corrupted admin profile stats
- negative post counts
- incorrect author profile metrics

Expected behavior:

The system should decrement the deleted post author's `postsCount`, not the deleting moderator/admin.

---

## Steps to reproduce

1. User A publishes a post.
2. Admin B deletes User A's post.
3. Check both profiles.

Result:

Admin B loses post count.
User A remains unchanged.

---

## Environment (if relevant)

Backend: Node.js + MongoDB

Affected Area:

- moderation system
- post lifecycle
- user profile stats

Primary File:

backend/src/app/modules/post/post.service.ts

---

N/A — backend data corruption issue.

Doremon-tech-svg:
## Related issue
Closes #3259

## What this PR does
Fixes a data integrity bug in `deletePost()` where deleting a published story always decremented the **logged-in user's** `postsCount`, instead of the **original post author's** `postsCount`.

This only showed up when the deleting user was different from the author — i.e. when an admin or moderator deleted someone else's published post. In that case:
- The admin/moderator's own `postsCount` was incorrectly decremented (even though they never owned that post).
- The actual author's `postsCount` was left unchanged, leaving stale/incorrect profile stats.
- Repeated admin deletions could push an admin's `postsCount` into negative numbers.

### The fix
In `backend/src/app/modules/post/post.service.ts`, inside `deletePost()`:

```diff
  if (post.isPublished) {
    await User.findByIdAndUpdate(
-     user._id,
+     post.author,
      { $inc: { postsCount: -1 } }
    );
  }
```

`user._id` is the person performing the delete action (could be the author, an admin, or a super_admin). `post.author` is the actual owner of the content. The decrement should always target the latter.

No other logic was touched:
- Authorization checks earlier in the function still correctly use `user._id` (these determine *who is allowed* to delete, which is unrelated to *whose stats get updated*).
- `post.deletedBy = user._id` is untouched — that's correctly tracking who performed the moderation action, separate from the stats bug.
- Soft-delete behavior, bookmark cleanup, and the delete API response are all unchanged.

## How this was tested
Manually verified the reproduction steps from the issue, since `post.service.ts` currently has no existing automated test coverage:

1. User A registers, creates, and publishes a post (`postsCount: 1`).
2. Admin B deletes User A's post via the delete endpoint.
3. Checked both users directly in MongoDB:
   - User A's `postsCount` correctly drops to `0`.
   - Admin B's `postsCount` remains unchanged.

Also verified: an author deleting their own post still works as before (no regression there, since `user._id === post.author` in that case).

## Note on build/lint checks
`npm run build -w backend` (`tsc`) currently fails on `main` with 30 pre-existing TypeScript errors across 10 unrelated files (`newsletter.controller.ts`, `user.interface.ts`, `collab.socket.ts`, etc.). Confirmed via `git stash` that these errors exist identically with this fix removed — none involve `post.service.ts` or this change. Flagging so it isn't mistaken for something this PR introduced.

The root-level `npm run lint` similarly fails due to ~48 pre-existing ESLint/parsing errors in unrelated frontend files. Backend has no `lint` script configured.

## Screenshot
N/A — backend-only data fix, verified via manual MongoDB inspection (see "How this was tested" above).

## Checklist
- [x] Bug fix
- [x] Tested locally (manual reproduction steps, see above)
- [x] Self-reviewed changes
- [x] Related issue linked
- [x] Fixed incorrect decrement target (`user._id` → `post.author`)
- [x] Preserved authorization/moderation logic
- [x] Preserved soft-delete logic
- [x] No breaking changes introduced
- [x] No API, schema, or frontend changes

### DIFF
--- a/backend/src/app/modules/post/post.service.ts
+++ b/backend/src/app/modules/post/post.service.ts
@@ -506,7 +506,7 @@ const deletePost = async (postId: string, token: ITokenPayload) => {
 
   if (post.isPublished) {
     await User.findByIdAndUpdate(
-      user._id,
+      post.author,
       { $inc: { postsCount: -1 } }
     );
   }


ronisarkarexe:
pls give a star

PR #3346, opened 2026-06-16
fix: move rate limiter after CORS to prevent preflight quota consumption and remove dead Cookie from allowedHeaders

### LINKED ISSUE #3345: fix: rate limiter mounted before CORS causes preflight requests to consume rate limit quota and Cookie is dead config in allowedHeaders

### Bug 1— Rate limiter before CORS
The rate limiter is registered before the CORS middleware. Every OPTIONS
preflight a browser sends to negotiate CORS permissions is counted against
the 100 requests per 15 minutes window. A single-page application making
requests to many endpoints can exhaust a significant portion of the quota
on preflights before any real API call is made. When the limit is hit the
browser receives a 429 response on the preflight. Because the CORS
middleware never runs for that response, no Access-Control-Allow-Origin
header is included. The browser reports this as a CORS error rather than
a rate limit error, making the root cause very difficult to diagnose.

CORS must be registered first so preflight responses always succeed and
the rate limiter only applies to actual API requests.

### Bug 2— Cookie in allowedHeaders is dead config
Cookie is listed in the CORS allowedHeaders array. Cookie is a
browser-forbidden request header — browsers manage it internally and
never include it in CORS preflight Access-Control-Request-Headers or
allow JavaScript to set it manually. The entry has no effect on any
browser. Cookie transmission is already handled by credentials: true.
The dead entry misleads contributors into thinking the API accepts or
requires a manually set Cookie header.

### File
backend/src/app.ts

### Steps to reproduce

Steps to Reproduce — Bug 1 (Rate limiter before CORS):

Deploy the backend with the current middleware order (limiter before CORS)
From a browser, make rapid API calls to more than 100 different endpoints within 15 minutes
Observe that OPTIONS preflight requests consume the rate limit quota
Once the limit is hit, send any new cross-origin request
Observe: browser reports a CORS error (No 'Access-Control-Allow-Origin' header) instead of a 429 — root cause is impossible to diagnose from the browser

Steps to Reproduce — Bug 2 (Dead Cookie in allowedHeaders):

Open browser DevTools → Network tab
Make any cross-origin request that triggers a CORS preflight (OPTIONS)
Inspect the preflight request headers under Access-Control-Request-Headers
Observe: Cookie is never listed — browsers never include it regardless of the allowedHeaders config
Confirm the entry has zero effect on any request or response

### Environment (if relevant)

_No response_

_No response_

### Record

akhilmodi29:
## Summary
Two fixes in app.ts — one corrects middleware ordering so CORS
preflight requests are never rate-limited, the other removes a dead
and misleading entry from allowedHeaders.

## Bug 1— Rate limiter before CORS
Moved the rateLimit middleware registration to after the CORS block.
OPTIONS preflight requests now receive proper CORS headers regardless
of traffic volume. Rate limiting still applies to all subsequent
non-preflight requests as intended.

## Bug 2 — Dead Cookie in allowedHeaders
Removed "Cookie" from the allowedHeaders array. It is a browser-
forbidden header that can never appear in a CORS preflight and has
no effect. Cookie transmission is handled by credentials: true.

## Changes
- backend/src/app.ts

Closes #3345

### DIFF
--- a/backend/src/app.ts
+++ b/backend/src/app.ts
@@ -17,14 +17,6 @@ const app: Application = express();
 app.set("trust proxy", 1);
 app.use(helmet());
 
-const limiter = rateLimit({
-  windowMs: 15 * 60 * 1000,
-  max: 100,
-  message: "Too many requests, please try again later.",
-});
-
-app.use(limiter);
-
 const defaultCorsOrigins =
   process.env.NODE_ENV === "development"
     ? ["http://localhost:4001", "http://localhost:4002"]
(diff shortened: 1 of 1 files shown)


ronisarkarexe:
pls give a star

PR #3351, opened 2026-06-16
fix: verify JWT before blacklist lookup, use ApiError for revoked tokens, attach fresh DB user to req.user and reject non-Bearer schemes in auth middleware

### LINKED ISSUE #3350: fix: auth middleware queries blacklist before JWT verification, bypasses global error handler for revoked tokens, sets stale JWT payload as req.user and silently accepts non-Bearer auth schemes

### Bug 1 — Blacklist DB lookup before JWT verification
The middleware queries TokenBlacklist.findOne before verifying the JWT
signature. Any request with a forged, expired, or completely random
Bearer token triggers a database query before being rejected. An
attacker can flood the server with invalid tokens causing unnecessary
DB load. The cryptographic verification is free and synchronous —
it must run first to gate all subsequent DB operations.

### Bug 2— Blacklist rejection bypasses global error handler
When a blacklisted token is detected the middleware calls
res.status(401).json({ message: "..." }) directly. Every other error
path throws ApiError and calls next(err) which routes through
globalErrorHandler for consistent response shaping. The blacklist
rejection produces a raw JSON structure that does not match the
standardised error format returned by every other 401/403 response.

### Bug 3— req.user set to stale JWT payload
After fetching the user from the database the middleware discards the
fresh document and attaches the raw decoded JWT payload to req.user.
If the user's role, email, or name was updated after the token was
issued, controllers read stale data. The fresh user document is already
available and should be used.

### Bug 4 — Non-Bearer auth schemes silently accepted
The token extraction falls back to authHeader.trim() when the header
does not start with Bearer. A Basic auth header or any arbitrary string
is used as the JWT value, causing a verifyToken failure only after
unnecessary processing.

### File
backend/src/app/middleware/auth.middleware.ts

### Steps to reproduce

Steps to Reproduce:
1. Send POST /api/v1/story/branching with Authorization: Bearer invalidtoken
2. Observe a MongoDB query is made to TokenBlacklist before the JWT
   verification rejects the token
3. Expected: JWT signature verified first, DB only queried for valid tokens

Steps to Reproduce:
1. Add a valid token to the TokenBlacklist collection
2. Make any authenticated request with that token
3. Observe the response body structure differs from other 401 responses
4. Expected: same error shape as all other ApiError responses

Steps to Reproduce:
1. Log in and obtain a JWT token
2. Update the user's role in the database
3. Make an authenticated request with the old token
4. In the controller, read req.user.role — it reflects the old role

Steps to Reproduce:
1. Send any request with Authorization: Basic dXNlcjpwYXNz
2. Observe the middleware attempts to use the full header value as a JWT
3. Expected: immediate 401 rejection for non-Bearer schemes

### Environment (if relevant)

_No response_

_No response_

### Record

akhilmodi29:
## Summary
Four security and correctness fixes in auth.middleware.ts covering
middleware execution order, error response consistency, stale data
exposure, and auth scheme validation.

## Bug 1 — JWT verified before blacklist DB query
Moved JwtHelpers.verifyToken above the TokenBlacklist.findOne call.
Invalid or forged tokens are now rejected synchronously with no DB
cost. The blacklist is only queried after the signature is confirmed.

## Bug 2 — ApiError thrown for blacklisted tokens
Replaced res.status(401).json() with throw new ApiError so the
rejection flows through globalErrorHandler and produces the same
standardised response shape as all other auth errors.

## Bug 3 — Fresh DB user attached to req.user
Changed (req as any).user = verifiedUser to (req as any).user = user
so controllers always receive up-to-date profile data rather than
values frozen at token issuance time.

## Bug 4 — Non-Bearer schemes rejected explicitly
Replaced the silent fallback with an explicit check that throws 401
immediately when the Authorization header does not start with Bearer.

## Changes
- backend/src/app/middleware/auth.middleware.ts — all four fixes applied

Closes #3350

### DIFF
--- a/backend/src/app/middleware/auth.middleware.ts
+++ b/backend/src/app/middleware/auth.middleware.ts
@@ -77,9 +77,6 @@ const auth =
           );
         }
 
-        // Bug #111 Fix: Attach the fresh DB user to req.user instead of the
-        // raw JWT payload so controllers always get up-to-date role, email,
-        // and profile data rather than values frozen at token issuance time.
         (req as any).user = user;
 
         next();
--- a/backend/src/app/middleware/auth.middleware.ts
+++ b/backend/src/app/middleware/auth.middleware.ts
@@ -10,74 +10,82 @@ import { USER_STATUS } from "../../enums/user_status";
 
 const auth =
   (...requiredRole: string[]) =>
-  async (req: Request, res: Response, next: NextFunction) => {
-    try {
-      const authHeader = (req.headers.authorization || "") as string;
+    async (req: Request, res: Response, next: NextFunction) => {
+      try {
+        const authHeader = (req.headers.authorization || "") as string;
 
-      const token = authHeader.startsWith("Bearer ")
-        ? authHeader.slice(7).trim()
-        : authHeader.trim();
+        if (!authHeader.startsWith("Bearer ")) {
+          throw new ApiError(
+            httpStatus.UNAUTHORIZED,
+            "You are not authorized to access"
+          );
+        }
 
-      if (!token) {
-        throw new ApiError(
-          httpStatus.UNAUTHORIZED,
-          "You are not authorized to access"
+        const token = authHeader.slice(7).trim();
+
+        if (!token) {
+          throw new ApiError(
+            httpStatus.UNAUTHORIZED,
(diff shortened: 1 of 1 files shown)


ronisarkarexe:
pls give a star

PR #3358, opened 2026-06-16
Fix/topics not reset on new story

### LINKED ISSUE #3341: BUG: reset topics when new story is generated

### What happened?

What's wrong: When a new story is generated, the useEffect([stories]) resets auto-save refs but never resets topics state. Any custom #tags the user added, or defaults they toggled off, bleed into the next story session — so published posts get tagged with topics from the previous story.

### Steps to reproduce

The fix: Add setTopics(topicsData) inside the existing useEffect(() => {...}, [stories]) block, right after the auto-save ref resets.

### Environment (if relevant)

_No response_

_No response_

### Record

OmanshiRaj:
FIXES #3341 
## What this PR does
When a new story is generated, custom topics added and defaults toggled off
during the previous session were never cleared — they bled into the new story.
Added `setTopics(topicsData)` inside the `useEffect([stories])` block to reset
topics back to defaults on every new generation.

## Steps to reproduce
1. Generate a story, add a custom topic e.g. `#dragonlore`, deselect a default e.g. `#adventure`
2. Generate a second unrelated story
3. Observe — `#dragonlore` and deselected `#adventure` are still present
4. Publishing at this point tags the post incorrectly

## Fix
One line added in `useEffect([stories])`:
`setTopics(topicsData);`

## Type of change
- [x] Bug fix

## Checklist
- [x] Tested locally — topics reset correctly on each new generation
- [x] Self-reviewed
- [x] N/A — logic fix, no screenshot needed

### DIFF
--- a/frontend/src/components/stories/stories.view.component.tsx
+++ b/frontend/src/components/stories/stories.view.component.tsx
@@ -310,6 +310,7 @@ const [, setShowRemix] = useState<boolean>(false);
     lastSavedContentRef.current = "";
     hasSavedSessionRef.current = false;
     savedPostIdRef.current = null;
+    setTopics(topicsData);
   }, [stories]);
 
   useEffect(() => {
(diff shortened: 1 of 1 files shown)


ronisarkarexe:
pld give a star

PR #3413, opened 2026-06-17
fix : Duplicate .contact-submit-btn and .contact-btn-gradient

zairahussain27:
These selectors are defined once under the "Submitting State Custom Transitions" section and then redefined later under the "Submit button" section.

Having duplicate definitions for the same selectors increases maintenance complexity and can lead to inconsistencies when future changes are applied to one block but not the other. Some properties are duplicated while others differ slightly between the two definitions.

This makes the source harder to maintain and debug.


ronisarkarexe:
pls give a star

Issue #3486, opened 2026-06-18
Feat: Add 'Share story' button with copy-link and social sharing options


Akshita-2307:
### Is there an existing issue for this?
- [X] I have searched the existing issues

### Describe the feature
Users currently cannot easily share stories with others. Add a share button on the story view page that provides options to copy the story link to clipboard, share via social media (Twitter/X, WhatsApp), or generate a shareable preview card.

### Acceptance criteria (optional)
- [ ] Share button on story view page
- [ ] 'Copy link' option with success toast
- [ ] Social sharing for Twitter/X and WhatsApp
- [ ] Share modal/popover with all options
- [ ] Proper og:meta tags for link previews when shared

### Screenshots or mockups (if applicable)
A share icon button that opens a small popover with 'Copy link' and social media icons, similar to Medium or dev.to share features.

### Record
- [X] I agree to follow this project's Code of Conduct
- [X] I want to work on this issue

PR #3496, opened 2026-06-18
Feat: Add share story button with Web Share API and clipboard fallback (#3486)

Akshita-2307:
Summary: Adds a Share button to the StoryViewer component that uses the native Web Share API on supported devices and falls back to copying the URL to the clipboard. Includes a toast notification for clipboard fallback.


ronisarkarexe:
pls give a star

Issue #3487, opened 2026-06-18
BUG: AI story analysis panel shows stale results when selecting different stories


Akshita-2307:
### Is there an existing issue for this?
- [X] I have searched the existing issues

### What happened?
When a user views AI analysis for one story, then navigates to another story's analysis, the previous story's analysis results briefly flash before the new ones load. The analysis panel does not clear its state while loading new data.

### Steps to reproduce
1. Open Story A and view its AI analysis
2. Navigate to Story B and click AI analysis
3. Story A's analysis appears briefly before Story B's results load
4. If Story B's analysis is loading, the stale data causes confusion

### Environment (if relevant)
All browsers. Likely caused by not resetting analysis state when story ID changes.

### Screenshots or screen recording (if helpful)
N/A

### Record
- [X] I agree to follow this project's Code of Conduct
- [X] I want to work on this issue

PR #3497, opened 2026-06-18
Fix: Clear stale prompt analysis data when prompt changes (#3487)

Akshita-2307:
Summary: Fixes stale analysis data in PromptAnalysisCard by clearing cached results when the prompt text changes. Uses a ref to track previous prompt value and resets analysis state on change.


ronisarkarexe:
pls give a star

Issue #3483, opened 2026-06-18
BUG: Theme preference does not persist after page refresh on some pages


Akshita-2307:
### Is there an existing issue for this?
- [X] I have searched the existing existing issues

### What happened?
When a user switches to dark/light mode on certain pages (e.g., Profile, Settings, Contact Us), the theme resets to default after a page refresh. The theme preference is not consistently persisted across all routes.

### Steps to reproduce
1. Navigate to the Settings page
2. Toggle from dark to light mode (or vice versa)
3. Refresh the page
4. Theme resets to default instead of maintaining the selected preference

### Environment (if relevant)
All browsers. Occurs specifically on secondary pages, not the home page.

### Screenshots or screen recording (if helpful)
N/A

### Record
- [X] I agree to follow this project's Code of Conduct
- [X] I want to work on this issue

Akshita-2307:
This feature is already implemented. \	heme.context.tsx\ (\rontend/src/components/theme/theme.context.tsx\) persists the theme preference:
- Line 15-26: \getInitialTheme()\ reads from \localStorage.getItem('theme')\
- Line 45: \localStorage.setItem('theme', theme)\ saves on every theme change
- Also respects \prefers-color-scheme\ media query as fallback

PR #3536, opened 2026-06-18
fix: theme preference already persists via localStorage (#3483)

vaishnavi003-svg:
## What this PR does
Reviewed and confirmed that theme persistence is already correctly implemented in the codebase.

**Current Implementation:**
- `getInitialTheme()` reads from `localStorage` on app load
- `useEffect` saves theme to `localStorage` on every change
- Dark class applied to `document.documentElement`
- System preference fallback (`prefers-color-scheme`) when no stored theme
- Glow preference also persisted via `localStorage`
- `useTheme` hook available for all components

**Files Reviewed:**
- `frontend/src/components/theme/theme.context.tsx`

**Verification:**
```typescript
✅ localStorage.getItem('theme') - reads stored theme
✅ localStorage.setItem('theme', theme) - saves on change
✅ root.classList.toggle('dark', theme === 'dark') - applies class
✅ window.matchMedia('(prefers-color-scheme: dark)') - fallback


vaishnavi003-svg:
@ronisarkarexe 
please review
thank you

PR #3655, opened 2026-06-20
Investigate TTS voice filtering issueFix TTS voice filtering issue

muskanrana2903-svg:
## Problem

The story narration feature was selecting voices through a strict filtering process. In some browsers and devices, this resulted in limited voice options being available, leading to unclear or unnatural speech output.

## Root Cause

The existing logic returned only the filtered voices when matches were found. When the filtered set was too small, users could end up with poor-quality voice selections, affecting narration clarity.

## Solution

- Updated the voice filtering logic to handle cases where only a few matching voices are available.
- Added a fallback mechanism that expands the available voice pool when the filtered results are insufficient.
- Ensured voice selection remains flexible across different browsers and operating systems.
- Improved the likelihood of selecting clearer and more natural-sounding voices for story narration.

## Impact

- Better voice quality for story playback.
- More consistent behavior across devices and browsers.
- Reduced chances of users receiving low-quality or difficult-to-understand voices.

## Testing

- Verified that voice lists are populated correctly after filtering.
- Tested fallback behavior when limited matching voices are available.
- Confirmed that story narration still works as expected after the changes.

### DIFF
--- a/frontend/src/hooks/useSpeechSynthesis.ts
+++ b/frontend/src/hooks/useSpeechSynthesis.ts
@@ -95,9 +95,11 @@ const filterVoicesByGender = (
   const pattern = gender === "female" ? femalePattern : malePattern;
   const filtered = browserVoices.filter((voice) => pattern.test(voice.name));
 
-  return filtered.length > 0 ? filtered : browserVoices;
+ if (filtered.length < 3) {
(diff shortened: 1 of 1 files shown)


ronisarkarexe:
pls give a star

PR #3683, opened 2026-06-20
fix: catch duplicate-key error on bookmark create to handle race condition gracefully

### LINKED ISSUE #3682: Bug: toggleBookmark doesn't handle duplicate-key race condition, can return unhandled 500 and inflate bookmarksCount
Describe the Bug

toggleBookmark in backend/src/app/modules/bookmark/bookmark.service.ts uses a check-then-act pattern (findOne followed by create/delete) without handling the race condition between the two operations.

The Bookmark schema has a compound unique index on { userId, storyId }, which correctly prevents duplicate bookmark documents at the database level. However, the service code does not catch the resulting duplicate-key error (MongoDB error code 11000).

If two requests for the same bookmark toggle arrive close together (double-click, slow network retry, etc.):

Both requests run findOne and see no existing bookmark
Both attempt Bookmark.create(...)
One succeeds; the second throws an unhandled MongoServerError: E11000 duplicate key
This unhandled error propagates as a generic 500 instead of a graceful response
Additionally, Post.findByIdAndUpdate(post._id, { $inc: { bookmarksCount: 1 } }) for the failed request may still execute before the create() rejection is handled depending on ordering, inflating bookmarksCount relative to the actual number of bookmark documents

Affected File

backend/src/app/modules/bookmark/bookmark.service.ts

Steps to Reproduce

Open a story page while logged in
Rapidly double-click the bookmark button (or fire two concurrent requests to the toggle endpoint for the same story)
Observe an unhandled server error in one of the two requests, and/or bookmarksCount drifting from the real bookmark count over repeated rapid toggles

Expected Behavior

A duplicate-key error from the unique index should be caught and treated as "already bookmarked" rather than surfacing as a 500
bookmarksCount should never drift from the actual count of Bookmark documents for a story

Suggested Fix

Wrap the create call in a try/catch for the duplicate-key case, and only increment bookmarksCount after a confirmed successful create:

ts} else {
  try {
    await Bookmark.create({ userId: user._id, storyId: post._id });
    await Post.findByIdAndUpdate(post._id, { $inc: { bookmarksCount: 1 } });
    return { message: "Story bookmarked!", isBookmarked: true };
  } catch (err: any) {
    if (err?.code === 11000) {
      return { message: "Story already bookmarked!", isBookmarked: true };
    }
    throw err;
  }
}

sweetylearner-max:
## Related Issue
Closes #3682

## Description of Changes
Wrapped the `Bookmark.create()` call in `toggleBookmark` with a try/catch block. If a concurrent request already created the bookmark (triggering the unique compound index on `{ userId, storyId }`, error code `11000`), the function now returns a graceful "already bookmarked" response instead of letting the error propagate as an unhandled 500.

## Root Cause
The check-then-act pattern (`findOne` then `create`) is not atomic. Two near-simultaneous requests (double-click, retry) could both pass the `findOne` check before either `create` call resolves, causing the second `create` to fail on the unique index without being caught.

## Type of Change
- [x] Bug fix

## Testing
- Single bookmark toggle — works as before (bookmark/unbookmark)
- Simulated duplicate-key error path — returns graceful response instead of 500

### DIFF
--- a/backend/src/app/modules/bookmark/bookmark.service.ts
+++ b/backend/src/app/modules/bookmark/bookmark.service.ts
@@ -30,9 +30,16 @@ const toggleBookmark = async (storyId: string, token: ITokenPayload) => {
     await Post.findByIdAndUpdate(post._id, { $inc: { bookmarksCount: -1 } });
     return { message: "Bookmark removed", isBookmarked: false };
   } else {
-    await Bookmark.create({ userId: user._id, storyId: post._id });
-    await Post.findByIdAndUpdate(post._id, { $inc: { bookmarksCount: 1 } });
-    return { message: "Story bookmarked!", isBookmarked: true };
+    try {
+      await Bookmark.create({ userId: user._id, storyId: post._id });
+      await Post.findByIdAndUpdate(post._id, { $inc: { bookmarksCount: 1 } });
+      return { message: "Story bookmarked!", isBookmarked: true };
+    } catch (err: any) {
+      if (err?.code === 11000) {
+        return { message: "Story already bookmarked!", isBookmarked: true };
+      }
+      throw err;
+    }
   }
 };
 
@@ -163,3 +170,5 @@ export const BookmarkService = {
   checkBookmarkStatus,
   deleteBookmark,
 };
+
+


sweetylearner-max:
Note: The build failure in CI comes from a pre-existing syntax error in `src/app/modules/story_version/enhance_prompt.utils.ts` (line 108+, unterminated template literal). This is unrelated to this PR — the file was already broken on `main` before this branch was created, and this PR only touches `bookmark.service.ts`.

PR #3749, opened 2026-06-20
fix: validate page, limit, dateFrom, and dateTo query params in search endpoint

### LINKED ISSUE #3748: bug: search endpoint doesn't validate page/limit/dateFrom/dateTo query params
Describe the Bug

SearchController.search in backend/src/app/modules/search/search.controller.ts does not validate several query parameters before passing them downstream, which can cause silent incorrect results or unhandled errors.

1. Unvalidated dateFrom/dateTo

These are passed straight into new Date(dateFrom) / new Date(dateTo) inside search.service.ts with no format check. An invalid value (e.g. dateFrom=notadate) produces an Invalid Date object. MongoDB does not throw on this — it silently filters using an invalid value, so the date range filter quietly returns wrong (often empty) results with no indication to the client that the input was malformed.

2. Unvalidated page/limit

tspage: page ? parseInt(page, 10) : 1,
limit: limit ? Math.min(parseInt(limit, 10), 50) : 10,

If page or limit is a non-numeric string (e.g. page=abc), parseInt returns NaN. This NaN flows into .skip((page - 1) * limit) in searchStories, which Mongoose either throws on or handles unpredictably, rather than returning a clean 400.

Affected Files

backend/src/app/modules/search/search.controller.ts
backend/src/app/modules/search/search.service.ts

Steps to Reproduce

Call GET /search?q=story&page=abc
Observe a server error or unexpected behavior instead of a validation error
Call GET /search?q=story&dateFrom=notadate
Observe the date filter silently returns incorrect results with no error

Expected Behavior

page/limit should be validated as positive integers, falling back to defaults or returning a 400 if invalid
dateFrom/dateTo should be validated as parseable dates, returning a 400 with a clear message if not

Suggested Fix

tsconst parsedPage = page ? Number(page) : 1;
const parsedLimit = limit ? Number(limit) : 10;

if (!Number.isInteger(parsedPage) || parsedPage < 1) {
  return sendResponse(res, {
    statusCode: httpStatus.BAD_REQUEST,
    success: false,
    message: "page must be a positive integer",
    data: null,
  });
}

if (!Number.isInteger(parsedLimit) || parsedLimit < 1) {
  return sendResponse(res, {
    statusCode: httpStatus.BAD_REQUEST,
    success: false,
    message: "limit must be a positive integer",
    data: null,
  });
}

if (dateFrom && isNaN(new Date(dateFrom).getTime())) {
  return sendResponse(res, {
    statusCode: httpStatus.BAD_REQUEST,
    success: false,
    message: "dateFrom is not a valid date",
    data: null,
  });
}

if (dateTo && isNaN(new Date(dateTo).getTime())) {
  return sendResponse(res, {
    statusCode: httpStatus.BAD_REQUEST,
    success: false,
    message: "dateTo is not a valid date",
    data: null,
  });
}

sweetylearner-max:
## Related Issue
Closes #3748

## Description of Changes
Added validation for query parameters in `SearchController.search`:
- `page` and `limit` are now checked with `Number.isInteger()` and must be positive — invalid values (e.g. `page=abc`) now return a clean 400 instead of letting `NaN` flow into `.skip()`/`.limit()`
- `dateFrom` and `dateTo` are validated with `isNaN(new Date(...).getTime())` — invalid date strings now return a 400 instead of silently producing wrong/empty search results

## Root Cause
None of these query params were validated before being parsed/used. Non-numeric `page`/`limit` produced `NaN` flowing into MongoDB pagination, and invalid date strings created `Invalid Date` objects that MongoDB silently filtered on without error.

## Type of Change
- [x] Bug fix

## Testing
- `q` validation still works as before (existing tests pass)
- `page=abc` / `limit=abc` now returns 400 with clear message
- `dateFrom=notadate` now returns 400 with clear message
- Valid requests behave exactly as before

### DIFF
--- a/backend/src/app/modules/search/search.controller.ts
+++ b/backend/src/app/modules/search/search.controller.ts
@@ -25,13 +25,52 @@ const search = catchAsync(async (req: Request, res: Response) => {
     });
   }
 
+  const parsedPage = page ? Number(page) : 1;
+  const parsedLimit = limit ? Number(limit) : 10;
+
+  if (!Number.isInteger(parsedPage) || parsedPage < 1) {
+    return sendResponse(res, {
+      statusCode: httpStatus.BAD_REQUEST,
+      success: false,
+      message: "page must be a positive integer",
+      data: null,
+    });
+  }
+
+  if (!Number.isInteger(parsedLimit) || parsedLimit < 1) {
+    return sendResponse(res, {
+      statusCode: httpStatus.BAD_REQUEST,
+      success: false,
+      message: "limit must be a positive integer",
+      data: null,
+    });
+  }
+
+  if (dateFrom && isNaN(new Date(dateFrom).getTime())) {
+    return sendResponse(res, {
+      statusCode: httpStatus.BAD_REQUEST,
+      success: false,
+      message: "dateFrom is not a valid date",
+      data: null,
+    });
+  }
+
+  if (dateTo && isNaN(new Date(dateTo).getTime())) {
+    return sendResponse(res, {
+      statusCode: httpStatus.BAD_REQUEST,
+      success: false,
+      message: "dateTo is not a valid date",
+      data: null,
+    });
+  }
+
   const result = await SearchService.search({
     q,
     type: type as "story" | "user" | "tag" | "all",
     genre,
     sortBy: sortBy as "relevance" | "date" | "popularity",
-    page: page ? parseInt(page, 10) : 1,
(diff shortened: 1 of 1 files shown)


sweetylearner-max:
Note: The CI build failure is unrelated to this PR. It's caused by a pre-existing dependency issue in `frontend/package.json` — `y-quill@^1.2.0` is declared but the latest published version is `1.0.0`, so `pnpm install` fails on `main` regardless of this PR's changes.

PR #3751, opened 2026-06-20
fix: check buffer length before timingSafeEqual to prevent RangeError on malformed payment signature

### LINKED ISSUE #3750: Bug: verifyPayment crashes with RangeError on malformed signature length instead of returning 400
Describe the Bug

verifyPayment in backend/src/controllers/payment.controller.ts calls crypto.timingSafeEqual() directly on a client-supplied signature without first checking buffer lengths.

tsconst isSignatureValid = crypto.timingSafeEqual(
  Buffer.from(expectedSignature, "hex"),
  Buffer.from(razorpay_signature, "hex")
);

crypto.timingSafeEqual() throws a RangeError if the two buffers are not the same length — it does not return false. Since razorpay_signature comes straight from req.body with no length validation, an attacker can send a signature string of any length (e.g. an empty string, or a single character) and trigger an uncaught exception instead of receiving a clean "invalid signature" response.

This is in the payment verification path, so a malformed/short signature should fail cleanly with a 400 — not crash the handler.

Affected File

backend/src/controllers/payment.controller.ts

Steps to Reproduce

Call POST /payment/verify with a valid razorpay_order_id and razorpay_payment_id
Set razorpay_signature to a string shorter than the expected 64-character hex digest (e.g. "abc")
Observe an unhandled RangeError instead of a 400 "signature verification failed" response

Expected Behavior

A malformed or wrong-length signature should be rejected with a clean 400, just like any other invalid signature
The handler should never throw an uncaught exception based on attacker-controlled input length

Actual Behavior

crypto.timingSafeEqual() throws when buffer lengths differ, and nothing here guards against that before the call.

Suggested Fix

tsconst expectedBuffer = Buffer.from(expectedSignature, "hex");
const providedBuffer = Buffer.from(razorpay_signature, "hex");

const isSignatureValid =
  expectedBuffer.length === providedBuffer.length &&
  crypto.timingSafeEqual(expectedBuffer, providedBuffer);

if (!isSignatureValid) {
  res.status(400).json({
    success: false,
    error: "Payment signature verification failed. This request may be tampered.",
  });
  return;
}

sweetylearner-max:
## Related Issue
Closes #3750

## Description of Changes
Added a length comparison before calling `crypto.timingSafeEqual()` in `verifyPayment`. The two buffers are now confirmed to be the same length before the constant-time comparison runs.

## Root Cause
`crypto.timingSafeEqual()` throws a `RangeError` when the two buffers it compares are not the same length. Since `razorpay_signature` is taken directly from `req.body` with no prior length validation, an attacker could send a malformed/short signature and trigger an uncaught exception in a payment-critical code path instead of a clean 400 response.

## Type of Change
- [x] Bug fix (security)

## Testing
- Valid signature of correct length — verifies as before
- Malformed/short signature — now returns 400 cleanly instead of throwing
- Tampered signature of correct length — still correctly rejected

### DIFF
--- a/backend/src/controllers/payment.controller.ts
+++ b/backend/src/controllers/payment.controller.ts
@@ -113,10 +113,12 @@ export const verifyPayment = async (
       .update(`${razorpay_order_id}|${razorpay_payment_id}`)
       .digest("hex");
 
-    const isSignatureValid = crypto.timingSafeEqual(
-      Buffer.from(expectedSignature, "hex"),
-      Buffer.from(razorpay_signature, "hex")
-    );
+    const expectedBuffer = Buffer.from(expectedSignature, "hex");
+    const providedBuffer = Buffer.from(razorpay_signature, "hex");
+
+    const isSignatureValid =
+      expectedBuffer.length === providedBuffer.length &&
+      crypto.timingSafeEqual(expectedBuffer, providedBuffer);
 
     if (!isSignatureValid) {
       res.status(400).json({


sweetylearner-max:
Note: The CI build and Type Check failures are caused by the same pre-existing `y-quill@^1.2.0` dependency resolution issue on `main` (the latest published version is `1.0.0`), unrelated to this PR's changes.

PR #3762, opened 2026-06-20
Fix/story parser locations 1035

vaishnavi003-svg:
## What this PR does
Fixed `parseStory()` in `storyParser.ts` to extract ALL occurrences of location words instead of only the first occurrence.

**Problem:**
- `indexOf()` only returned the first occurrence of each location word
- Location nodes missed important narrative context from later scenes
- Story graph was incomplete

**Solution:**
- Replaced `indexOf` with `matchAll` to find ALL occurrences
- Added context scoring system to pick the most relevant occurrence
- Added `occurrenceCount` to track total mentions

**Key Changes:**
```typescript
// ❌ Before: Only first occurrence
const idx = lowerContent.indexOf(word);

// ✅ After: All occurrences with context scoring
const positions = getAllOccurrences(content, word);
// Score each occurrence and pick the best one


vaishnavi003-svg:
@ronisarkarexe 
please review
thank you

vaishnavi003-svg:
@ronisarkarexe
please review
thank you

PR #3870, opened 2026-06-21
fix: return 400 for invalid MongoDB ObjectId (#3833)

### LINKED ISSUE #3833: fix : return 400 for invalid MongoDB ObjectId in handleCastError
Summary of What Needs to be Done:
handleCastError currently returns HTTP status code 500 when a MongoDB CastError occurs (e.g., when an invalid ObjectId format is passed as a parameter). This is incorrect because an invalid ID format is a client-side mistake (malformed input), not a server-side failure. The correct HTTP status for this case is 400 (Bad Request).

Changes that Need to be Made:
- backend/src/errors/handle_cast_error.ts: Change statusCode from 500 to 400 in the return object

Impact that it would Provide:
- Correct HTTP semantics: 4xx errors indicate client mistakes, 5xx indicate server failures
- Better API consumer experience: clients can distinguish between bad input (400) and internal errors (500)
- Aligns handleCastError with handleZodError which already uses 400

Note: Please assign this issue to the `tmdeveloper007` account.

saurabhhhcodes:
## Fixes #3833

Changed `statusCode` from 500 to 400 in `handleCastError` since an invalid ObjectId format is a client-side mistake (malformed input), not a server-side failure.

- backend/src/errors/handle_cast_error.ts: statusCode 500 → 400

### DIFF
--- a/backend/src/errors/handle_cast_error.ts
+++ b/backend/src/errors/handle_cast_error.ts
@@ -2,7 +2,7 @@ import mongoose from "mongoose";
 import { IGenericErrorMessage } from "../interfaces/error";
 
 const handleCastError = (err: mongoose.Error.CastError) => {
-  const statusCode = 500;
+  const statusCode = 400;
   const errors: IGenericErrorMessage[] = [
     {
       path: err.path,


ronisarkarexe:
pls give a star

PR #3886, opened 2026-06-21
Fix duplicate React tree render in main.tsx

### LINKED ISSUE #3876: Fix duplicate React tree render in main.tsx (App renders twice)
## Issue
In \rontend/src/main.tsx\, the entire React tree (\GoogleOAuthProvider > Provider > ThemeProvider > App\) is rendered twice inside \<StrictMode>\. This causes duplicate providers and double-rendering of the entire app.

## Fix
Remove the first duplicate block (lines 25-34 with commented-out HelmetProvider) or the second duplicate block (lines 35-41). Only one copy of the provider tree should remain.

## Severity
High - causes double API calls, double event listeners, potential state issues.

sanrishi:
Removes the duplicate React provider tree in main.tsx that was causing the app to render twice.

Closes #3876

### DIFF
--- a/frontend/src/main.tsx
+++ b/frontend/src/main.tsx
@@ -23,15 +23,6 @@ if (!container) {
 
 createRoot(container).render(
   <StrictMode>
-    {/* <HelmetProvider> */}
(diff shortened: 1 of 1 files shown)


ronisarkarexe:
pls give a star

PR #3889, opened 2026-06-21
Remove stale tmp/ directory with 5 upstream files

### LINKED ISSUE #3882: Remove stale tmp/ directory with 5 upstream files
The tmp/ directory at the project root contains 5 stale upstream files that are not used by the application. These are leftover temporary copies that should be cleaned up. Fix: delete the entire tmp/ directory.

sanrishi:
Deletes the stale tmp/ directory containing 5 upstream/temporary files that are not used by the application.

Closes #3882


sanrishi:
@ronisarkarexe You can review this!

PR #3920, opened 2026-06-22
fix: add expired token and invalid iat/exp order checks in auth-validator

### LINKED ISSUE #3919: [BUG] auth-validator.ts accepts expired and temporally invalid JWT tokens
Describe the Bug

`validateTokenPayload` in `frontend/src/utils/auth-validator.ts` validates the `exp` and `iat` claims for type and format, but misses two critical checks:

1. **Expired token not detected** — if `exp` is in the past, the token is accepted as valid instead of being rejected.
2. **Impossible timestamps not detected** — if `iat >= exp` (issued-at is the same as or after expiry), this logically invalid/corrupted token is not caught.

Steps to Reproduce

1. Pass a decoded JWT payload where `exp` is a past timestamp (e.g. `exp: 1000`)
2. `validateTokenPayload` will pass without throwing — expired token treated as valid
3. Similarly, pass `iat >= exp` — no error is thrown

Suggested Fix

After the existing `exp` check, add:

const currentTime = Math.floor(Date.now() / 1000);
if (decodedData.exp < currentTime) {
  throw new Error("Token has expired.");
}
if (decodedData.iat >= decodedData.exp) {
  throw new Error("Token 'iat' must be before 'exp'.");
}

sweetylearner-max:
Fixes #3919

## Bug
`validateTokenPayload` in `frontend/src/utils/auth-validator.ts` was missing two critical JWT checks:
1. Expired tokens were accepted as valid (no `exp` vs current time check)
2. Logically invalid tokens where `iat >= exp` were not caught

## Fix
- Added expiry check: compares `exp` against `Math.floor(Date.now() / 1000)`
- Added timestamp order check: throws if `iat >= exp`

Contributing under GSSoC'26.

### DIFF
--- a/frontend/src/utils/auth-validator.ts
+++ b/frontend/src/utils/auth-validator.ts
@@ -1,17 +1,11 @@
-/**
- * Validates a decoded JWT token payload and strictly checks its claims format and existence.
- * Throws a descriptive Error if validation fails.
- */
-export const validateTokenPayload = (decodedData: any): void => {
+export const validateTokenPayload = (decodedData: Record<string, unknown>): void => {
   if (!decodedData || typeof decodedData !== "object") {
     throw new Error("Token payload is not a valid object.");
   }
-  // Support _id, userId, or sub for the user identifier claim
   const userId = decodedData.userId || decodedData._id || decodedData.sub;
   if (!userId || typeof userId !== "string" || userId.trim() === "") {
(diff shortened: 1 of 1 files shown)


sweetylearner-max:
Hi @ronisarkarexe! PR #3920 is ready for review whenever you get a chance. All checks are passing. Thank you!

PR #3948, opened 2026-06-22
Fix: Validate AI provider API keys at startup (#3761)

### LINKED ISSUE #3761: [SECURITY] API key configuration not validated before AI calls
## Description

The AI API key is not validated before making story generation calls. If the key is missing or invalid, the API call will fail with a confusing error message. Worse, if the key is accidentally committed, it's exposed.

## Suggested Fix
Validate the API key at startup and provide a clear configuration error if missing. Ensure the API key is loaded from environment variables only, never hardcoded.

Tharsiga-21:
Closes #3761 — Validate AI provider API keys at startup

Problem

The AI API key was never validated before making story generation calls. Each provider client (getOpenAIClient, getGeminiClient, getAnthropicClient in ai.service.ts) only checked for its key lazily, the first time it was actually called — deep inside a story generation request. This meant:

If no AI provider key was configured at all, the app would start normally and only fail later with a confusing error, triggered by whichever user happened to request a story first.
There was no single, clear startup-time signal telling a developer "you haven't configured any AI provider."

Fix

Added a validateAIProviderKeys() check in backend/src/config/index.ts, run once at startup (right after config is loaded, before the server starts accepting requests):

If none of OPEN_AI_KEY / OPENAI_API_KEY, GEMINI_API_KEY, or ANTHROPIC_API_KEY are set, the app now fails fast with a clear, actionable error:

  Error: No AI provider API key configured. Set at least one of OPEN_AI_KEY, GEMINI_API_KEY, or ANTHROPIC_API_KEY in your environment. See backend/.env.example for setup instructions.

If some (but not all) provider keys are missing, the app starts normally but logs a clear warning per missing provider, e.g.:

  [Config] OPEN_AI_KEY not set — OpenAI provider unavailable.
  [Config] ANTHROPIC_API_KEY not set — Anthropic provider unavailable.

This matches the existing multi-provider fallback design in ai.service.ts (OpenAI → Gemini → Anthropic), since the app is still fully functional as long as one provider is configured.

This follows the same pattern already used for JWT_SECRET / JWT_REFRESH_SECRET (requiredEnv()), extending startup validation to cover the AI provider keys too.

Screenshot 2026-06-22 170745

Also fixes (pre-existing, unrelated syntax errors blocking local dev)

While testing this fix locally, the backend failed to even compile due to three pre-existing syntax errors already present on main, unrelated to this issue:

backend/src/app/modules/ai_model/ai_model.utils.ts — a stray, unmatched ); inside the .map() callback for normalizing storyboard scenes, breaking the function body.
backend/src/utils/promptSecurity.ts — appears to be leftover damage from a bad merge: an orphaned, incomplete canonicalizeSecurityText function (never closed, missing .trim() and a leftover main marker line) tangled together with a second working implementation (normalizeInput). Both validateAndFormatPrompt and validateOutput contained two interleaved, duplicate implementations. Cleaned up to keep the complete, working version of each function only; confirmed canonicalizeSecurityText had no other usages in the codebase before removing it.

These were blocking npm run dev entirely (TypeScript/esbuild compile failures), independent of the API key validation work. Fixed them so the app could actually boot and the new validation logic could be tested end-to-end.

Testing

Confirmed the app throws the new startup error when zero AI provider keys are set (screenshot attached).
Confirmed the app starts normally with warnings when only GEMINI_API_KEY is set, leaving OPEN_AI_KEY and ANTHROPIC_API_KEY unset (screenshot attached). For this test, GEMINI_API_KEY was set to a placeholder string (not a real key) locally, purely to verify the presence check passes — no actual call to the Gemini API was made or needed to confirm this validation logic works.
Confirmed no AI API key is hardcoded anywhere — all keys are read from environment variables only, consistent with existing config/index.ts patterns.
Verified npm run dev compiles and boots cleanly after the syntax fixes above (remaining ECONNREFUSED errors for MongoDB/Redis are expected locally, since no local DB/Redis instance is running — unrelated to this change).

Files changed

backend/src/config/index.ts — added validateAIProviderKeys()
backend/src/app/modules/ai_model/ai_model.utils.ts — removed stray syntax error
backend/src/utils/promptSecurity.ts — resolved leftover merge conflict damage

Note for maintainer

The two files listed under "Also fixes" above were already broken on main before I started — I didn't introduce these issues, I just hit them as blockers while testing locally. Happy to split these into a separate PR if you'd prefer to review/merge them independently from the API key validation fix — just let me know.


Tharsiga-21:
The Type Check / typecheck CI check is failing, but the remaining errors are pre-existing and unrelated to this PR — they're already present on main:

src/app/modules/collab/collab.service.ts(15,13): error TS2352 — Binary/Buffer type mismatch
src/app/modules/collab/yjs.gateway.ts(3,26): error TS2307 — missing 'lodash' type declarations
src/app/modules/collab/yjs.gateway.ts(17,5): error TS2740 — Socket.IO Namespace/Server type mismatch
src/controllers/character.controller.ts(2,27): error TS2307 — missing module '../models/Character.model'

None of these touch files changed in this PR. I confirmed this by running tsc locally before and after my changes — these four errors are present either way, so main currently fails typecheck for any PR, not just this one.

For reference, this PR did fix six TS2304: Cannot find name 'sanitizeJsonText' errors that were introduced by my changes (a function that existed locally in storyBranchingController.ts but wasn't exported/imported where ai_model.utils.ts needed it) — those are resolved in the latest commit.

Happy to open a separate issue for the four collab/character errors above if that's helpful, but didn't want to fix them as part of this PR since they're outside its scope and I don't have context on the intended behavior there.

Issue #3997, opened 2026-06-22
[BUG] Redis queue initialization crashes application startup when REDIS_URL is not configured


Daksh7785:
# [BUG] Redis queue initialization crashes application startup when REDIS_URL is not configured

**Type:** Bug
**Severity:** High
**Priority:** High
**Component:** Story Request Queue Service
**File:** `backend/src/services/storyRequestQueue.ts`

## Description

The story request queue service imports and initializes Redis dependencies at module load time.

The service relies on:

```ts
import Redis from "ioredis";
```

and attempts to create a Redis connection even when:

```env
REDIS_URL
```

is not configured.

In non-Redis development environments, `ioredis` defaults to attempting a connection against:

```text
localhost:6379
```

If Redis is not running locally, connection attempts fail during application startup.

This prevents contributors from running the application despite Redis-backed queueing not being required for all development workflows.

---

### Startup Flow

```text
Application Startup
        │
        ▼
Import storyRequestQueue.ts
        │
        ▼
Instantiate Redis Client
        │
        ▼
REDIS_URL Missing
        │
        ▼
Attempt localhost:6379
        │
        ▼
Connection Error
        │
        ▼
Process Crash
```

### Example Error

```text
Error: connect ECONNREFUSED 127.0.0.1:6379
```

or

```text
MaxRetriesPerRequestError
```

depending on Redis configuration.

---

### Local Development Failure

A contributor performs:

```bash
npm install
npm run dev
```

without Redis installed.

Result:

```text
Application fails to start.
```

Even if the queue system is never used.

---

### Docker Development Failure

A developer may intentionally run:

```bash
docker-compose up
```

without a Redis service.

The backend immediately crashes.

---

### Poor Onboarding Experience

New contributors encounter infrastructure errors before reaching the application.

This increases setup friction and slows development.

---

### 1. Startup Instability

Missing Redis configuration can prevent the entire backend from booting.

### 2. Reduced Contributor Experience

Developers must install and configure Redis before they can work on unrelated features.

### 3. Environment Coupling

The queue system becomes a hard dependency even when queue functionality is unused.

### 4. CI/Test Failures

Automated test environments without Redis may fail unexpectedly.

### 5. Local Development Friction

Simple development tasks require additional infrastructure.

---

## Root Cause

Redis initialization occurs unconditionally.

Conceptually:

```ts
const redis = new Redis(process.env.REDIS_URL);
```

or equivalent logic executes during module import.

When:

```env
[...]

PR #4036, opened 2026-06-22
fix: avoid Redis startup crash without REDIS_URL

saurabhhhcodes:
Summary:\n- Make the shared Redis client optional when REDIS_URL is missing\n- Reuse the safe client in token quota checks so the backend can boot without Redis\n- Add a regression test for the no-Redis path\n\nTesting:\n- npm test -- --runInBand src/__tests__/rateLimitStore.test.ts\n- git diff --check\n\nCloses #3997


ronisarkarexe:
pls give a star

PR #4049, opened 2026-06-22
Fix explore search state

Tenisha5132:
## Screenshot (when helpful)

N/A – This change only affects state reset logic and does not introduce any UI modifications.

### Problem

When a user performed a search on the Explore page, navigated away, and later returned to the Explore page, the previous search query and results remained visible. Instead of loading the default Explore feed, the component retained stale search state from the previous session.

### Solution

Updated the `useEffect` hook in `ExploreComponent` to depend on `location.key` obtained from React Router's `useLocation()` hook. This ensures that every navigation event leading back to the Explore page triggers a reset of the local search-related state, including:

* `searchTerm`
* `selectedTags`
* `page`
* Search results and related filters

By resetting these values whenever the Explore page is re-entered, users consistently see the default Explore feed rather than stale search results.

## Type of change

* [x] Bug fix
* [ ] New feature
* [ ] Code refactor
* [ ] Documentation update

## Related issue

None

N/A

## Checklist

* [x] I have added screenshots or a recording when they help explain this PR, or noted **N/A** with a one-line reason.
* [x] I have filled in the related issue number when there is one.
* [x] I have tested my changes by performing searches, navigating away from the Explore page, and returning to verify that the default feed loads correctly.
* [x] I have done a self-review of the code.


ronisarkarexe:
done push package lock file

PR #4256, opened 2026-06-25
fix: remove token blacklist race condition in auth middleware

laxmiprasanna188:
### Changes made

* Removed MongoDB `TokenBlacklist` lookup from authentication middleware.
* Updated authentication flow to rely on existing JWT `tokenVersion` validation.
* Removed per-request blacklist database query to improve authentication performance.

### Reason

The previous implementation checked token blacklist status separately, which could create a timing gap during concurrent logout/token revocation scenarios.

Using `tokenVersion` validation ensures revoked tokens are invalidated at the authentication layer without requiring a database blacklist lookup on every request.

### Testing

* Verified valid JWT tokens continue to authenticate.
* Verified invalid token versions are rejected.
* Verified role and account status checks remain unchanged.

### DIFF
--- a/backend/src/app/middleware/auth.middleware.ts
+++ b/backend/src/app/middleware/auth.middleware.ts
@@ -5,7 +5,6 @@ import { Secret } from "jsonwebtoken";
 import ApiError from "../../errors/api_error";
 import { JwtHelpers } from "../../utils/jwt.helper";
 import { User } from "../modules/user/user.model";
-import { TokenBlacklist } from "../modules/auth/tokenBlacklist.model";
 import { USER_STATUS } from "../../enums/user_status";
 
 type JwtVerifiedUser = {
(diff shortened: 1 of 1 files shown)


laxmiprasanna188:
Hi @ronisarkarexe 

I checked the failing CI checks.

The PR changes only:
`backend/src/app/middleware/auth.middleware.ts`

The failing build/typecheck errors are coming from existing syntax issues in:

* `backend/src/app/modules/ai_model/ai_model.utils.ts`
* `backend/src/utils/promptSecurity.ts`

These files are not modified in this PR.

The implemented fix:

* Removed TokenBlacklist database lookup from auth middleware
* Relied on existing JWT tokenVersion validation
* Removed per-request blacklist query overhead

Please let me know if any further changes are required.

PR #4270, opened 2026-06-25
fix : removed duplicate reject call in raceGenerationWithTimeout catch block

tmdeveloper007:
Closes #4265.

Removed the unconditional reject(error) call after the if/else branch in the catch handler of raceGenerationWithTimeout. Previously, when the operation failed with a non-abort error, reject was called twice: once in the else branch and once after. This caused an unhandled promise rejection because the promise was already settled.

Changes Made:
- backend/src/utils/generation_timeout.ts: Removed the trailing reject(error) after the if/else block in the catch handler.

Impact it Made:
- Prevents unhandled promise rejections on genuine operation errors.
- Makes the error path consistent: timeout errors resolve to GenerationTimeoutError, genuine errors propagate normally.

### DIFF
--- a/backend/src/utils/generation_timeout.ts
+++ b/backend/src/utils/generation_timeout.ts
@@ -88,7 +88,6 @@ export const raceGenerationWithTimeout = async <T>(
           controller.abort();
           reject(error);
         }
-        reject(error);
       });
   });
 };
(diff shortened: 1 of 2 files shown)


tmdeveloper007:

PR #4350, opened 2026-06-26
fix : added typeof window guard to loadRazorpayScript for SSR safety

tmdeveloper007:
Closes #4345.

Added a typeof window === 'undefined' guard at the top of the loadRazorpayScript function to prevent runtime crashes when the utility is imported in server-side rendering environments such as Next.js or Remix.

Changes Made:
- frontend/src/utils/loadRazorpay.ts: Added SSR guard that returns Promise.resolve(false) immediately when running outside a browser context.

Impact it Made:
- Prevents SSR crashes when loadRazorpayScript is called on the server
- Makes the utility safe for universal/isomorphic React applications
- Consistent with other localStorage-using utilities in the codebase that already have typeof window guards


tmdeveloper007:

PR #4375, opened 2026-06-27
fix: replace any type with proper ApiError interface in 404 handler

prajwal-2509:
Fixes #4368
Replaced const error: any with a properly typed ApiError interface in the 404 handler, so the error shape attached via next(error) is now type-checked instead of bypassing TypeScript entirely.
Note: npm run build currently fails on unrelated pre-existing errors in ai_model.utils.ts and promptSecurity.ts (not touched by this PR). Verified this specific file compiles cleanly via npx tsc --noEmit backend/src/app.ts.
I'd like to work on this — contributing under GSSoC'26


prajwal-2509:
Is the PR is merged?

PR #4398, opened 2026-06-27
fix : fixed useScrollDirection infinite re-render by using ref for lastScrollY

tmdeveloper007:
Closes #4393.

Fixed the useScrollDirection hook by replacing useState for lastScrollY with useRef. The previous implementation used lastScrollY in the useEffect dependency array, causing an infinite loop because setLastScrollY was called inside the effect body, triggering re-renders on every scroll event.

Changes Made:
- frontend/src/hooks/useScrollDirection.ts: Changed lastScrollY from useState to useRef, removed lastScrollY from useEffect dependency array

Impact it Made:
- Fixes infinite render loop causing performance issues on pages using useScrollDirection
- ESLint and TypeScript checks pass locally


tmdeveloper007:

PR #4417, opened 2026-06-27
fix: remove duplicate Published Stories menu item in dashboard.utils.ts

### LINKED ISSUE #2926: BUG: Duplicate "Published Stories" menu item in dashboard.utils.ts causes redundant sidebar entry
Is there an existing issue for this?

No, I have searched the existing issues

What happened?
In frontend/src/components/dashboard/dashboard.utils.ts, the menuItems array contains two separate entries for "Published Stories" pointing to the same path /dashboard/published-stories:

Entry at index 1: name: "Published Stories", icon: "fas fa-book", roles: [USER, WRITER]
Entry at index 3: name: "Published Stories", icon: "fas fa-book-open", roles: [USER, WRITER, ADMIN, SUPER_ADMIN]

This means users with USER or WRITER roles will see the "Published Stories" link twice in their sidebar.
Steps to Reproduce

Log in as a user with the USER or WRITER role
Open the dashboard sidebar
Observe "Published Stories" appears twice

Expected Behavior
Only one "Published Stories" entry should appear in the sidebar.
Suggested Fix
Remove the first duplicate entry (index 1) and update the second entry's roles to include all four roles: USER, WRITER, ADMIN, SUPER_ADMIN.

harshap0:
Fixes #2926

## What changed?
- Removed duplicate "Published Stories" menu item from dashboard.utils.ts
- Kept single entry with fas fa-book-open icon covering all roles: USER, WRITER, ADMIN, SUPER_ADMIN

## Why?
Users with USER or WRITER roles were seeing "Published Stories" twice in the sidebar.

### DIFF
--- a/frontend/src/components/dashboard/dashboard.utils.ts
+++ b/frontend/src/components/dashboard/dashboard.utils.ts
@@ -30,6 +30,7 @@ export const menuItems: MenuItem[] = [
     ],
   },
   {
+     // Fix #2926: Removed duplicate "Published Stories" entry; unified roles to include all four roles
     name: "Published Stories",
     icon: "fas fa-book-open",
     path: "/dashboard/published-stories",


harshap0:
After re-checking, the duplicate "Published Stories" entry from #2926 doesn't actually exist in current main — only one entry is present with the correct roles. My change here was just a comment, not a real fix. Closing this PR — will look for a genuinely reproducible issue instead.

PR #4440, opened 2026-06-28
fix: persist OpenAI's temporary image url server-side (#4284)

adrish-mage:
## Before you open this PR
- **Issue:** Fixes #4284
- **Description:** OpenAI's `url` response is a temporary, expiring link. This PR fetches the image bytes server-side immediately after generation and converts them into a persistent `data:` URI, so storyboard images never silently break after the link expires.

## Screenshot (when helpful)
All 5 unit tests passing locally:

```
PASS src/utils/__tests__/storyboard_image_generation.test.ts
  generateStoryboardImage
    ✓ fetches the temporary OpenAI url server-side and returns a persisted data URI instead of the raw url (#4284)
    ✓ returns null (mapped to imageStatus: failed by the caller) when the temporary url can no longer be fetched
    ✓ leaves the b64_json branch untouched, since it never returns a temporary link
    ✓ returns null when neither url nor b64_json is present
    ✓ returns null without calling fetch when no provider is configured

Test Suites: 1 passed, 1 total
Tests:       5 passed, 5 total
```

## What this PR does
OpenAI's `url` response field is a short-lived, OpenAI-hosted CDN link. Previously it was saved as-is into `imageUrl`, so storyboard images silently broke once the link expired with no recovery path.

This PR adds a `persistTemporaryImageUrl` helper in `storyboard_image_generation.ts` that fetches the image bytes from the temporary URL immediately after generation while the link is still valid  and converts them into a self-contained `data:` base64 URI. This mirrors how the existing `b64_json` branch already behaves and was never affected by this bug. No schema or caller changes are needed since `imageUrl` already accepts data URIs.

If the re-fetch fails (e.g. URL already expired), the function returns `null`, which the existing caller in `story_visualizer.service.ts` already maps to `imageStatus: "failed"`  so no scene is ever left holding a broken URL.

## Type of change
- [x] Bug fix

## Related issue
Fixes #4284

N/A : this is a backend-only bug fix with no UI changes; test output above serves as proof.

## Checklist

### DIFF
--- a/backend/src/utils/__tests__/storyboard_image_generation.test.ts
+++ b/backend/src/utils/__tests__/storyboard_image_generation.test.ts
@@ -1,0 +1,141 @@
+describe("generateStoryboardImage", () => {
+  const OLD_ENV = process.env;
+  const ORIGINAL_FETCH = global.fetch;
+
+  beforeEach(() => {
+    jest.resetModules();
+    process.env = { ...OLD_ENV };
+    process.env.IMAGE_GENERATION_PROVIDER = "openai";
+    process.env.IMAGE_GENERATION_API_KEY = "test-api-key";
+  });
+
+  afterEach(() => {
+    global.fetch = ORIGINAL_FETCH;
+  });
+
+  afterAll(() => {
+    process.env = OLD_ENV;
+  });
+
+  const mockGenerationResponse = (body: unknown) => {
+    return {
+      ok: true,
+      json: async () => body,
+    } as Response;
+  };
+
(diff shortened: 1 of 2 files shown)


adrish-mage:
The 2 failing CI checks (CI/build and Type Check/typecheck) are pre-existing 
failures on main caused by syntax errors in ai_model.utils.ts and promptSecurity.ts, 
unrelated to this PR. These failures exist on main before this branch was created.

adrish-mage:
Hey @ronisarkarexe Do let me know if changes are needed in this PR.

PR #4444, opened 2026-06-28
fix: remove duplicate routes and nested duplicate subtrees in App.tsx

Harshada-Yele:
## Screenshot (when helpful)

N/A — routing structure change, verified via manual 
navigation testing (listed below), no UI changes.

## What this PR does

App.tsx had extensive route duplication from what appears 
to be a bad merge:

1. ~18 routes were declared twice as siblings in the same 
   children array (once eager, once lazyPage-wrapped) — in 
   React Router v6 only the first sibling match is ever used, 
   so the second copy was dead code.
2. The /dashboard route's children contained a full duplicate 
   copy of nearly all public routes that don't belong there.
3. /auth/email-validation and the payment/collab ProtectedRoute 
   block were each duplicated (top-level AND nested in /dashboard).
4. The actual DashboardLayout (profile, settings, writers, 
   users, analytics, post-lists) was nested one level deeper 
   than necessary.

Removed all duplicates, kept the lazyPage-wrapped version of each route (consistent with the rest of the file's code-splitting pattern), and flattened DashboardLayout to be a direct child of /dashboard.

## Complexity
**High-:**
Required reading the entire routing tree, understanding React Router v6's first-match-wins sibling behavior, and tracing which nested ProtectedRoute/role combinations were genuinely needed vs accidental duplicates.

## Testing
Manually verified every route category renders correctly:
- Public routes (home, login, signup, about-us, etc.)
- Protected routes redirect to /login when logged out 
  (stories, collab, payment)
- Role-gated dashboard routes (users/writers = admin-only, 
  analytics = writer-only, post-lists = writer+admin)
- 404 catch-all still works
- No console errors on any route

## Type of change

- [x] Bug fix
- [ ] New feature
- [x] Code refactor
- [ ] Documentation update

## Related issue

Fixes #4443 

N/A

## Checklist


Harshada-Yele:
Hi @ronisarkarexe 
PR Ready for Review under Gssoc26

Issue #4642, opened 2026-07-01
BUG: Circular Self-Rendering and JSX Syntax Error in Pricing Component


Harshithk951:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### What happened?

In `frontend/src/components/home/pricing/pricing.component.tsx`, the component attempts to import itself (`import PricingComponent from "./pricing.component"`) and render `<PricingComponent />` inside `PricingMainComponent`. This causes a JSX syntax error (`JSX expressions must have one parent element`) and would result in infinite rendering/recursion at runtime.

### Steps to reproduce

1. Run `pnpm --dir frontend exec tsc -b --noEmit`
2. Observe JSX and typecheck errors for `frontend/src/components/home/pricing/pricing.component.tsx`.

### Environment (if relevant)

Frontend pricing component.

### Screenshots or screen recording (if helpful)

N/A

### Record

- [x] I agree to follow this project's Code of Conduct
- [x] I want to work on this issue

PR #4644, opened 2026-07-01
fix(frontend): restore pricing component and fix blog closing tag

Harshithk951:
### Description

This PR resolves pricing component circular self-rendering loops, JSX syntax errors, and mismatched closing tags in the frontend workspace.

### Proposed Changes

1. **pricing.component.tsx** (`frontend/src/components/home/pricing/pricing.component.tsx`):
   - Restored the full pricing page implementation (which was accidentally deleted in commit `47ce2483`).
   - Removed circular self-import/rendering of `PricingComponent` that would cause infinite loops and recursion at runtime.
   - Removed duplicate `highlight: true` key in the Pro plan object literal.

2. **blog.tsx** (`frontend/src/components/footer/blog.tsx`):
   - Replaced mismatched closing tag `</main>` on line 149 with `</div>` to match the outer container opening `div`.

### Verification

Validated locally by executing targeted typechecks on both modified files. Both files compile successfully with zero errors.


Harshithk951:
Closes #4642
Closes #4643

PR #4854, opened 2026-07-05
fix: deduplicate CORS origins configuration in server.ts and app.ts

sahitya0xsingh:
## Summary

Fixes **#4849** — Duplicate CORS origins configuration in `server.ts` and `app.ts`.

## Root Cause

The `defaultCorsOrigins` array was hardcoded in both `app.ts` and `server.ts`. This created a dual source of truth, making the codebase prone to configuration drift and potential CORS misconfigurations if the two definitions ever diverged.

## Changes Made

**`backend/src/app.ts`**

- Exported `defaultCorsOrigins` so it can be reused across the application.

**`backend/src/server.ts`**

- Removed the hardcoded `defaultCorsOrigins` definition.
- Imported `defaultCorsOrigins` directly from `app.ts` to establish a single source of truth:
  ```ts
  import app, { defaultCorsOrigins } from "./app";
  ```

## Testing

- [x] Verified that `defaultCorsOrigins` is accurately sourced from `app.ts`
- [x] Verified that the backend compiles and starts without CORS configuration errors
- [x] Verified that development origins (`http://localhost:4001`, etc.) are still correctly applied in development mode

## Related

Closes #4849

### DIFF
--- a/backend/src/app.ts
+++ b/backend/src/app.ts
@@ -18,7 +18,7 @@ const app: Application = express();
 app.set("trust proxy", 1);
 app.use(helmet());
 
-const defaultCorsOrigins =
+export const defaultCorsOrigins =
   process.env.NODE_ENV === "development"
     ? ["http://localhost:4001", "http://localhost:4002"]
     : ["https://storysparkai.vercel.app"];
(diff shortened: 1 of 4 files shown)


ronisarkarexe:
pls give a star

PR #4874, opened 2026-07-05
fix: invalidate JWT tokens on password change via passwordChangedAt

anmoltrivedi2312-ux:
Closes #4356

## Problem
Previously issued JWT tokens remained valid after a password change, since the JWT itself doesn't know its issuing user's password was updated. A stolen token could keep authenticating until natural expiry, even after the user reset their password.

## Fix
- Added `passwordChangedAt` to `IUser` interface and the Mongoose schema.
- Set `passwordChangedAt` in the `pre("save")` hook whenever an existing user's password is modified (with a 1s buffer to avoid same-second edge cases).
- In `auth.middleware.ts`, compare the JWT's `iat` claim against `passwordChangedAt`; reject the request with 401 if the token was issued before the password change.

## Testing
- Verified `changePassword` and `resetPassword` in `auth.service.ts` both mutate the Mongoose document and call `.save()`, so the `pre("save")` hook (and thus `passwordChangedAt`) fires correctly for both flows.
- Logic manually traced through the middleware to confirm token rejection when `iat < passwordChangedAt`.
- Full end-to-end manual test (login → change password → reuse old token → expect 401) not yet run against a live server — happy to verify further if a maintainer requests it.

## Files changed
- `backend/src/app/modules/user/user.interface.ts`
- `backend/src/app/modules/user/user.model.ts`
- `backend/src/app/middleware/auth.middleware.ts`


anmoltrivedi2312-ux:
Note: I noticed upstream's main no longer includes the role-authorization check in auth.middleware.ts. I've kept it during this merge since removing it looked like it could be an unintentional regression rather than a deliberate change happy to remove it if that was intentional.

anmoltrivedi2312-ux:
Also please add gssoc approved labels.

PR #5000, opened 2026-07-12
test(shared): add unit tests for send_response utility

Tharsiga-21:
## What
Adds unit tests for `send_response.ts`, which is used across all backend
endpoints to standardize the JSON response shape (`success`, `statusCode`,
`message`, `meta`, `data`).

## Tests added
- sends correct JSON structure with all fields
- preserves the correct HTTP status code
- handles response with meta fields (page, limit, total)
- handles response with null data
- uses null for message when not provided

Express's `Response` is mocked (`status`/`json` as jest mocks returning
`this`) so no real server is needed.

## Note on file location
The linked issue specifies `backend/src/shared/tests/send_response.test.ts`,
but this repo's `jest.config.js` only matches `**/__tests__/**/*.test.ts` —
a plain `tests/` folder is never picked up by the test runner. Placed the
file at `backend/src/shared/__tests__/send_response.test.ts` instead, next
to the existing `pick.test.ts`, so it's actually discovered.

## Verification
All 5 tests pass locally against the current implementation.

Closes #4572

### DIFF
--- a/backend/src/shared/__tests__/send_response.test.ts
+++ b/backend/src/shared/__tests__/send_response.test.ts
@@ -1,0 +1,99 @@
+import { Response } from "express";
+import sendResponse from "../send_response";
+
+const makeRes = (): Partial<Response> => ({
+  status: jest.fn().mockReturnThis(),
+  json: jest.fn().mockReturnThis(),
+});
+
+describe("send_response", () => {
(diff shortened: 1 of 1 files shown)


Tharsiga-21:
hi @ronisarkarexe  a pr is linked for the issue #4572 for review

PR #6788, opened 2026-08-17
fix: use correct tag type for cache invalidation in writer application API

### LINKED ISSUE #5749: BUG: Writer application form remains available after a successful submission
## Is there an existing issue for this?

## What happened?

After a writer application succeeds, the profile query is not invalidated. The writer-application mutation uses `invalidatesTags: ["User"]`, but the registered and provided tag is lower-case `user`. The cached profile therefore keeps `isApplyForWriter: false`, so the profile continues to show the application form instead of the “Application Under Review” state until the user reloads or refetches their profile.

## Steps to reproduce

1. Sign in as a non-writer account with `isApplyForWriter` set to false.
2. Open the profile page and submit the Writer Application form successfully.
3. Stay on the page after the success toast appears.
4. Observe that the application form remains visible rather than changing to the pending-review message.

## Environment (if relevant)

Frontend profile page; RTK Query cache invalidation.

Not applicable; the mismatch is in the mutation tag and the cached profile state.

## Record

aditya226-sharma:
The writer application mutation used `"User"` (uppercase) for `invalidatesTags` but the registered tag type is `"user"` (lowercase). This caused RTK Query to not invalidate the user profile cache after a successful writer application submission.

Fixes #5749

### DIFF
--- a/frontend/src/redux/apis/writer_application.api.ts
+++ b/frontend/src/redux/apis/writer_application.api.ts
@@ -1,4 +1,5 @@
 import baseApi from "../base_api/base.api";
+import { tagTypes } from "../tag-types";
 
 export const writerApplicationApi = baseApi.injectEndpoints({
   endpoints: (build) => ({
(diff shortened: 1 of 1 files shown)


aditya226-sharma:
@ronisarkarexe Gentle ping — ready for review!

